PR #67, opened 2015-03-19
Fix add truncated time point with Sunday as day 0

### LINKED ISSUE #65: Normal time point + out of bounds truncated time point causes a loop forever
A user made the assumption that, given Monday was -1, Sunday would be -0 (indexes starting from 0):

W-0T23/P1W

Where used in cylc, this seems to result in an infinite loop of trying to evaluate it.

This sort of thing should error out/be rejected.

matthewrmshin:
The logic was stuck in a `while` loop when adding a truncated time point like
`-W-0` to a normal time point.

Fix #65.

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -1160,6 +1160,9 @@ class TimePoint(object):
                 new._tick_over()
         if day_of_week is not None:
             new.to_week_date()
+            new._tick_over()
+            if new.day_of_week > 0 and day_of_week == 0:
+                day_of_week = Calendar.DAYS_IN_WEEK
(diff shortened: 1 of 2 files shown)


matthewrmshin:
@benfitzpatrick please review and check correctness.

benfitzpatrick:
I agree with @arjclark in the issue that it should "error out/be rejected".

matthewrmshin:
In which case, I'll modify the original issue to be more generic. The actual problem is that `TimePointParser.parse` and `TimePoint.__init__` do not check bounds.

PR #74, opened 2017-06-09
Slot datetime classes

oliver-sanders:
Makes datetime objects more memory efficient.

Also fixes an attribute which may have been set erroneously.


matthewrmshin:
@dvalters please sanity check.

PR #75, opened 2017-07-10
Fix quotes.

hjoliver:
Fix quotes in `parsers.py`


matthewrmshin:
1 review is OK with this change.

PR #78, opened 2017-08-09
Fix pylint errors and warnings

matthewrmshin:
Should eliminate all errors and most warnings.

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -63,6 +63,7 @@ class Calendar(object):
 
     @classmethod
     def default(cls):
+        """Return the singleton instance. Create if necessary."""
         if cls._DEFAULT is None:
             cls._DEFAULT = cls()
         return cls._DEFAULT
@@ -242,7 +243,7 @@ class TimeRecurrence(object):
         next_timepoint = timepoint + self.duration
         if self._get_is_in_bounds(next_timepoint):
             return next_timepoint
-        if (self.format_number == 1 and next_timepoint > self.end_point):
+        if self.format_number == 1 and next_timepoint > self.end_point:
             diff = next_timepoint - self.end_point
             if (2 * diff < self.duration and
                     self._get_is_in_bounds(self.end_point)):
@@ -317,11 +318,6 @@ class TimeRecurrence(object):
         elif self.format_number == 4:
             return prefix + str(self.duration) + "/" + str(self.end_point)
         return "R/?/?"
-
-    def get_tests(self):
-        """Return a series of self-tests."""
-        for recur_expression, result_points in self.TEST_EXPRESSIONS:
-            yield recur_expression, result_points
 
 
 class Duration(object):
@@ -448,7 +444,7 @@ class Duration(object):
 
     def get_is_in_weeks(self):
         """Return whether we are in week representation."""
-        return (self.weeks is not None)
+        return self.weeks is not None
 
     def to_days(self):
         """Convert to day representation rather than weeks."""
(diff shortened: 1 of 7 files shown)


isodatetime/data.py
```diff
@@ -1703,12 +1700,11 @@ def get_is_leap_year(year):
 
 def get_days_in_year_range(start_year, end_year):
     """Return the number of days within this year range (inclusive)."""
-    return _get_days_in_year_range(start_year, end_year,
-                                   calendar_mode=CALENDAR.mode)
+    return _get_days_in_year_range(start_year, end_year, CALENDAR.mode)
```
oliver-sanders: Do we still need the `CALENDAR.mode` argument?

isodatetime/data.py
```diff
@@ -1048,10 +1044,10 @@ def apply_time_zone_offset(self, offset):
                     offset.minutes / float(CALENDAR.MINUTES_IN_HOUR))
             else:
                 self.minute_of_hour += offset.minutes
-            self._tick_over()
+            self.tick_over()
```
oliver-sanders: Why have some methods changed from private to public?

isodatetime/parser_spec.py
```diff
@@ -285,7 +285,13 @@ def __str__(self):
 
 
 def get_date_translate_info(num_expanded_year_digits=2):
-    expanded_year_digit_regex = "\d" * num_expanded_year_digits
+    """Return list of 4-element tuples with date translate information.
+
+    Each tuple is (
+        basic regex, regex with capture group, format substitution,
+        property name)
+    """
```
oliver-sanders: Would be good if we could stick to a standard docstring format e.g: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

I.E. something like...

```
returns:
    list: List of four element tuples with the following components.
        - basic regex (str) - Some description text
        - regex with capture group (str) - ...
        ...
```

isodatetime/data.py
```diff
@@ -1390,7 +1385,7 @@ def _add_months(self, num_months):
             if self.get_is_week_date():
                 was_week_date = True
             self.to_calendar_date()
-        for i in range(abs(num_months)):
+        for _ in range(abs(num_months)):
```
dvalters: I get why PyLint suggests this, but I find it less readable/intuitive. (Oh well)

isodatetime/data.py
```diff
@@ -1750,11 +1746,11 @@ def _get_days_in_year_range(start_year, end_year, calendar_mode=None):
 
 def get_days_in_year(year):
     """Return the number of days in this particular year."""
-    return _get_days_in_year(year, calendar_mode=CALENDAR.mode)
+    return _get_days_in_year(year, CALENDAR.mode)
 
 
 @util.cache_results
-def _get_days_in_year(year, calendar_mode=None):
+def _get_days_in_year(year, _):
```
dvalters: If the argument isn't used, why does it need a 'discardable' argument passing? (I must be missing something here...)

matthewrmshin: It is used by `util.cache_results`.

isodatetime/data.py
matthewrmshin: It is used by `util.cache_results`.

isodatetime/data.py
matthewrmshin: In `TimePoint.add_truncated`, the method is invoked as `new.tick_over`, which is a bit bad according to pylint.

isodatetime/parser_spec.py
matthewrmshin: Done.

PR #79, opened 2017-08-10
More pylint related changes

matthewrmshin:
This should remove all but 1 remaining issues raised by Codacy.

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -215,9 +215,7 @@ class TimeRecurrence(object):
                     self.end_point - self.duration * (self.repetitions - 1))
         else:
             raise BadInputError(
-                BadInputError.RECURRENCE,
-                (i[:2] for i in inputs)
-            )
+                BadInputError.RECURRENCE, [i[:2] for i in inputs])
 
     def get_is_valid(self, timepoint):
         """Return whether the timepoint is valid for this recurrence."""
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -205,22 +205,18 @@ class TimeRecurrence(object):
             # Third form.
             self.format_number = 3
             if self.repetitions is not None:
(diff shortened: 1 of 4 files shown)


isodatetime/tests.py
```diff
@@ -1404,8 +1401,7 @@ def test_timepoint_at_year(test_year):
         test_data += data.get_days_since_1_ad(year - 1)
         assert_equal(test_data, ctrl_data)
         for attribute, attr_max in test_duration_attributes:
-            delta_attr = random.randrange(0, attr_max)
-            kwargs = {attribute: delta_attr}
+            kwargs = {attribute: random.randrange(0, attr_max)}
```
oliver-sanders: Thank you codacy-bot, luckily this pseudo-random generator is not being used for either security of cryptographic purposes.

isodatetime/data.py
```diff
@@ -205,22 +205,18 @@ def __init__(self, repetitions=None, start_point=None,
...
-                for i in range(self.repetitions - 1):
-                    point -= self.duration
-                self.start_point = point
+                self.start_point = (
+                    self.end_point - self.duration * (self.repetitions - 1))
         else:
             raise BadInputError(
                 BadInputError.RECURRENCE,
-                [i[:2] for i in inputs]
+                (i[:2] for i in inputs)
```
oliver-sanders: Unfortunately this is one case where a generator must be evaluated.

```python
>>> raise ValueError('here', (i for i in range(10)))
ValueError: ('here', <generator object <genexpr> at 0x7f2689502a00>)
>>> raise ValueError('here', [i for i in range(10)])
ValueError: ('here', [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```

matthewrmshin: Reverted.

PR #81, opened 2017-08-17
Fix time point addition.

### LINKED ISSUE #80: Day of the month inconsistency with truncated time points
When adding a truncated `TimePoint` to a regular `TimePoint` there can be inconsistencies with the day of the month.

e.g: `2000-01-01T00:00:00Z + 03-30T00:00:00Z = 2000-03-02T00:00:00Z`

Reported in https://github.com/cylc/cylc/issues/2382

oliver-sanders:
Closes #80 

This behaviour was caused by the `_tick_over` method correctly identifying that there is no 30th of February and rolling the date over to the 1st of March. This typically correct behaviour is not applicable to adding truncated time points as the day of the month needs to be carried over intact so this PR disables it.

The consequence of doing so is that if you try to add an invalid date isodatetime will yield an invalid date, e.g:

`2000-01-01T00:00:00Z + 02-30`
before: `2001-02-01T00:00:00Z`
after: `2001-02-30T00:00:00Z`

__PR for discussion__, is this simple logic change sufficient?

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -1186,9 +1186,11 @@ class TimePoint(object):
                 new.tick_over()
         if month_of_year is not None:
             new.to_calendar_date()
+            new.day_of_month = 0  # tick_over month independent of day.
             while new.month_of_year != month_of_year:
                 new.month_of_year += 1
                 new.tick_over()
+            new.day_of_month = day_of_month
         if year_of_decade is not None:
             new.to_calendar_date()
             new_year_of_decade = new.year % 10
--- a/isodatetime/tests.py
+++ b/isodatetime/tests.py
@@ -689,6 +689,30 @@ def get_timepoint_subtract_tests():
     ]
 
 
+def get_timepoint_add_truncated_tests():
+    """Yields tests for adding truncated timepoints to regular timepoints."""
+    return [
+        (
+            {"year": 2000, "month_of_year": 1, "day_of_month": 2,
+             "hour_of_day": 0, "minute_of_hour": 0, "second_of_minute": 0,
+             "time_zone_hour": 0, "time_zone_minute": 0},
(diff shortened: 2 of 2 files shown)


dvalters:
@oliver-sanders shall I review this as well (since I'm doing #79 too)

dvalters:
Not yet sure why, but this breaks one of the unit tests in `cylc/time_parser.py`. Specifically:

```python
    def test_inter_cycle_timepoints(self):
        """Test the inter-cycle point parsing."""
        task_cycle_time = self._parsers[0].parse_timepoint("20000101T00Z")
        tests = [("T06", "20000101T0600Z", 0),
                 ("-PT6H", "19991231T1800Z", 0),
                 ("+P5Y2M", "20050301T0000Z", 0),
 >>>>            ("0229T", "20000229T0000Z", 0),           # Now gives 20010229T0000Z (??)
                 ("+P54D", "20000224T0000Z", 0),
                 ("T12+P5W", "20000205T1200Z", 0),
                 ("-P1Y", "19990101T0000Z", 0),
                 ("-9999990101T00Z", "-9999990101T0000Z", 2),
                 ("20050601T2359+0200", "20050601T2159Z", 0)]
        for expression, ctrl_data, num_expanded_year_digits in tests:
            parser = self._parsers[num_expanded_year_digits]
            test_data = str(parser.parse_timepoint(
                expression,
                context_point=task_cycle_time))
            self.assertEqual(test_data, ctrl_data)
```

dvalters:
It doesn't seem to like Februaries for some reason with that unit test. (I thought it was maybe a leap year thing at first, but if you try any other date in February (e.g. `0201T`, `0217T` etc.), it rolls the year over to 2001. The other months work fine though...(??)

matthewrmshin:
Agreed with @oliver-sanders that we need a better fix.

PR #82, opened 2017-11-02
Rationalise duration output when subtracting timepoints 

benfitzpatrick:
Rationalise subtracting a later timepoint from an earlier one.
The previous logic would have:
```
20171102T0000Z - 20171102T0600Z => P-1DT18H
```

now produces `-PT6H`.

This is likely to be slower than the original logic, since it checks for A > B each time - the impact on e.g. cylc needs to be evaluated.

The new test with the negative duration (`-PT1H`) does not pass on current master - it produces `P-1DT23H`!

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -1324,6 +1324,8 @@ class TimePoint(object):
 
     def __sub__(self, other):
         if isinstance(other, TimePoint):
+            if other > self:
+                return -1* (other - self)
             other = other.copy()
             other.set_time_zone(self.get_time_zone())
             my_year, my_day_of_year = self.get_ordinal_date()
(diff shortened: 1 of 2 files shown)


matthewrmshin:
(Test failing due to pep8 => pycodestyle issue.)

oliver-sanders:
pycodestyle issue fixed -> re-running test battery.

oliver-sanders:
pycodestyle error: `isodatetime/data.py:1328:26: E225 missing whitespace around operator`

benfitzpatrick:
> pycodestyle error: isodatetime/data.py:1328:26: E225 missing whitespace around operator

Rubbish! Fixed.

Issue #89, opened 2018-09-23
Not able to run isodatetime tests under different time zone


kinow:
Hi,

I was trying to build and run the tests of isodatetime, but running the tests as in Travis CI (or even with `python -m unittest -v isodatetime.tests`) result in:

```python
Test the recurring date/time series parsing. ... ok

======================================================================
FAIL: test_timepoint_strftime_strptime (__main__.TestSuite)
Test the strftime/strptime for date/time expressions.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/kinow/.local/lib/python2.7/site-packages/isodatetime/tests.py", line 1388, in test_timepoint_strftime_strptime
    self.assertEqual(test_data, ctrl_data, strftime_string)
  File "/home/kinow/.local/lib/python2.7/site-packages/isodatetime/tests.py", line 1027, in assertEqual
    super(TestSuite, self).assertEqual(test, control, superinfo)
AssertionError: Source %m :?foobar++(%j%d %s:%M %H?foobar%F :%Y++(%X %S: produced:
'03 :?foobar++(06001 1014985802:30 12?foobar2002-03-01 :2002++(12:30:02 02:'
should be:
'03 :?foobar++(06001 1014939002:30 12?foobar2002-03-01 :2002++(12:30:02 02:'

----------------------------------------------------------------------
Ran 17 tests in 103.576s

FAILED (failures=1)
```

(error in %s, one time got 10149**858**02, and the other 10149**390**02)

Debugging the test, I think it was in the `Duration#get_days_and_seconds` that I realized it was getting the time for one day before. It means that the time that was 1st of March 2002 was being treated as NZ time, and then when retrieving the seconds, it would first convert the time to UTC, resulting in Feb 28th, time around 11:30 AM.

Tried some monkey patching, or re-defining the timezone in the command line with no luck. Even though users could be asked to set the TZ to UTC when running tests, I think it would be harder/frustrating to some users. Also, writing code for isodatetime forces me to fiddle with my TZ in order to run all the tests.

kinow:
Found a way to run the tests (thanks to co-worker Wolfgang). Just had to use

```
TZ=UTC python -m isodatetime.tests
```

Closing this ticket :tada:

PR #90, opened 2018-09-23
Use decimal for seconds instead of octal

kinow:
Not able to run the tests locally (see #89) so hoping Travis CI will be happy with this change :tada: 

I think the test was supposed to use a decimal and not an octal, but if it is the other way, then I would update this PR to use the 3.x syntax, with `0o2` instead.

Cheers
Bruno


kinow:
Oh, and that's from an error I got when trying to use it with Python 3 BTW. Hoping to have some more time to experiment with isodatetime more with Python 3 in the future.

Thanks!

PR #92, opened 2018-09-25
Report test coverage

kinow:
In #87 there were a few comments about adding support to Python 3 (while perhaps dropping support to Python 2). So as in Cylc and Rose (still broken, but will revisit it later) this pull request adds coverage reports to the project.

The build time increased significantly due to the way that Coverage.py instruments via Python's trace. The test `test_timerecurrence_parser` which normally takes a while, unfortunately takes a bit more due to the coverage overhead.

I executed the job around 10 times between yesterday and today, locally and in Travis-CI, and the current configuration runs all tests between 28-34 minutes. Here's the latest job output in Travis-CI for my fork.

https://travis-ci.org/kinow/isodatetime

Which is configured with m Codacy integration project API, uploading to my own dashboard for now

https://app.codacy.com/project/kinow/isodatetime/dashboard

Overall number for code coverage for isodatetime right now is an impressive **86%** :clap: :clap: :clap: 

But there are a few files that I would write some tests before porting to Python 3, like

- `timezone.py` has only about half of its code tested - https://app.codacy.com/app/kinow/isodatetime/file/22958614608/coverage
- The __str__ method of most classes are not tested. Easy to cover with a unit test, and even though it's using `.format()` which is less likely to cause issues doing a jump from Py2 to Py3, I think it's still worth writing a short test just to confirm that's not changed after the port
- In `data.py`, the `Duration#to_weeks` method is using a division, which could be different in Py3. So worth covering
- Also in `data.py`, `TimePoint#get` has some division not covered by tests
- And the `TimePoint` class has also some methods for addition with little coverage
- Still in `TimePoint`, now the `__cmp__` method uses `cmp`, which doesn't exist in Py3 if I remember well. Better cover 100% of this method as we could re-use the tests even after breaking that method into multiple methods - if necessary
- Not sure what `_type_checker` does or how it's used, but it appears to be easy to add a few more tests to avoid one of the branches introducing a regression

So maybe before moving the code to Python 3, we could write a few more tests, compare the coverage and issues in Codacy, and then start introducing the changes?

Cheers
Bruno

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------------
+# (C) British Crown Copyright 2013-2018 Met Office.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+
+# Codecov settings
+# After modifying this file, it might be worth to validate it with:
+# `curl --data-binary @.codecov.yml https://codecov.io/validate`
+
+# which branch to monitor and whether wait for ci success
+codecov:
+  branch: master
+  notify:
+    require_ci_to_pass: yes
+
+# define the colour bar limits here
+coverage:
+  precision: 2
+  round: down
+  range: "75...100"
+
+  # diff type
+  status:
+    project:
+      default:
+        # commits below this threshold will be marked as failed
+        target: '75%'
+        # how much we allow the coverage to drop
+        threshold: '5%'
+    patch:
+      default:
+        # basic
+        target: '75%'
+        threshold: '5%'
+        base: auto
+        # advanced
+        branches: null
+        if_no_uploads: error
+        if_not_found: success
+        if_ci_failed: error
+        only_pulls: false
+        flags: null
+        paths: null
+    changes: no
+
+# files to ignore
+ignore:
+  - "isodatetime/tests.py"
+
+# turn off comments to pull requests
+comment: off
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,48 @@
 
 ---
 language: python
+
 python:
-    - "2.6"
-    - "2.7"
-install:
-    - pip install pycodestyle
+  - 2.6
+  - 2.7
+env:
+  - RUN_COVERAGE=false
+  - RUN_COVERAGE=true
+
+matrix:
+  fast_finish: true
+  exclude:
+    - python: 2.6
+      env: RUN_COVERAGE=true
+    - python: 2.7
+      env: RUN_COVERAGE=false
+
+before_install:
+  # Coverage dependencies
+  - pip install coverage pytest-cov
+  # Code style dependencies
+  - pip install pycodestyle
 script:
-    - python -m isodatetime.tests
-    - pycodestyle -v isodatetime
+  - |
+    export COVERAGE_PROCESS_START="${TRAVIS_BUILD_DIR}/.coveragerc"
+    # NOTE: the way to handle long-living commands in the documentation is
+    # travis_wait 40 coverage run -m isodatetime.tests
+    # but it does not produce any output, which can be frustrating in case of errors.
+    # So here we are using the approach documented in the issue
+    # https://github.com/travis-ci/travis-ci/issues/4190.
+    coverage run -m isodatetime.tests &
+    minutes=0
+    limit=40
+    while kill -0 $! >/dev/null 2>&1; do
+      echo -n -e " \b" # never leave evidences!
+      if [ $minutes == $limit ]; then
+        break;
+      fi
+      minutes=$((minutes+1))
+      sleep 60
+    done
+after_script:
+  - coverage xml --ignore-errors
+  # Report metrics, such as coverage
+  - '[[ "$RUN_COVERAGE" = true ]] && bash <(curl -s https://codecov.io/bash)'
+  - pycodestyle -v isodatetime
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+# ----------------------------------------------------------------------------
+# (C) British Crown Copyright 2013-2018 Met Office.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# ----------------------------------------------------------------------------
+
+
+[run]
+
+branch=True
+cover_pylib=False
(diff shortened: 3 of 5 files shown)


kinow:
Last comment today, promise :-) Travis will very likely fail. That's due to the fact that the last stage, `report`, tries to upload the Coverage metrics to Codacy. Unless Travis-CI settings for `metomi/isodatetime` contain the Project API configured (CODACY_PROJECT_TOKEN = (secret)).

But once added, it should work and successfully upload the coverage - see https://support.codacy.com/hc/en-us/articles/207279819-Coverage for more

matthewrmshin:
Token added. Let's see if it works or not.

kinow:
Thanks @matthewrmshin if you have added the token to Travis-CI (making sure that option to display the value is not checked, so anonymous users are not able to use it), then perhaps you would like to retry the build too (there should be a button somewhere in https://travis-ci.org/metomi/isodatetime) :grin:

kinow:
Pushed a new commit to add the badge for later, but it failed too :cry: from the output, looks like it was not able to find the token.

matthewrmshin:
That's annoying. I have checked twice and the token is definitely added to the settings. Is there anything else that needs to be done?

kinow:
Not really @matthewrmshin 

After enabling Codacy for my fork of isodatetime, I copied the Project API from my Codacy

step1

Then I also had to enable analysis and make my branch as the main one (but we can skip this one for isodatetime, as we will be using `master`, and it probably has the analysis enabled already).

step2

Finally, in Travis-CI, I added the Project API (from the first step) as an environment variable, not displayed to users in the Web UI, but available during the build.

step3

The variable needs to be exactly `CODACY_PROJECT_TOKEN`.

kinow:
I am now looking at the output of my latest successful job in Travis-CI, and the job for this pull request, and I think I just had a Today_I_Learned moment. Looks like secret variables are always disabled in Travis-CI for pull requests.

Found it after the message in the failed job for this pull request, that says

>Encrypted environment variables have been removed for security reasons.
>See https://docs.travis-ci.com/user/pull-requests/#Pull-Requests-and-Security-Restrictions

And indeed, they explain why they remove it. Preventing coverage of running in pull requests is not very hard, but I think we are using Coveralls to measure the coverage of pull requests in some Apache projects (so that in some cases we can ask contributors to add tests when Coveralls tells us that the coverage went down).

I think you've done everything correct in Codacy & Travis-CI @matthewrmshin . Let me see if there's another way to test it.

kinow:
Well, couldn't find any other alternative, so asked the Codacy support. Let's wait a few days to see if they have any suggestions. Here's my request

>Hi, I am looking for a clarification on whether I can use Codacy and Travis-CI for pull requests. I am setting up an existing project, that is built on Travis-CI and is already sending metrics to Codacy, to now also send code coverage.
>
>However, it appears that as the CODACY_PROJECT_TOKEN is a secure environment variable, it gets disabled for pull requests for security reasons (e.g. a contributor could change .travis.yml and print that variable).
>
>I contribute to some projects that have coverage provided by Coveralls + Travis-CI, where Coveralls comments in the tickets with the coverage result for the pull request (e.g. https://github.com/apache/commons-lang/pull/275#issuecomment-395960195).
>
>Is there an alternative for the coverage of pull requests with Codacy + Travis-CI? Please.
>
>As we have contributors submitting pull requests from their own forks, it is not practical to build only from the main repository. And looking at the coverage after the pull requests have been merged is also not ideal.
>
>Thank you
>Bruno

kinow:
Had another look before calling it a day, and from the articles that I found, people were combining Travis-CI and Codacy, with another solution for coverage, like Codev.

[Setting up Codecov in Badgeyay](https://blog.fossasia.org/setting-up-codecov-in-badgeyay/)
[More tools: Travis CI, Codacy, and Codecov](https://www.implementingquantlib.com/2017/05/github-integrations.html)

Quoting from the first link

>BadgeYaY already has Travis CI and Codacy to test code quality and Pull Request but **there was no support for testing Code Coverage in repository against every Pull Request**.

So not much hope about having code coverage for each pull request with Codacy.

Would Codecov or Coveralls be OK as alternative? They are both free, so the only inconvenience would be having to log in and integrate another service with Travis-CI

matthewrmshin:
@hjoliver We adopted Codacy after you did for Cylc. Do you have any opinion on the other code coverage solutions that @kinow mentioned above?

hjoliver:
No - I have no opinion or knowledge on that!  @kinow can probably figure this out for us.

kinow:
Codacy support replied today, but just to acknowledge the ticket, and say they will reply with more details later 😐

kinow:
Last update from Codacy confirmed that it's not possible at the moment (but they said they will work on this feature)

>After checking with our developers your report we found no workaround that would protect the token. 

Codecov was the easiest to set up, so will update this pull request to use that instead.

kinow:
Looks like Codecov is working now!

The configuration for Codacy can be removed (that CODACY_PROJECT_TOKEN, there should be a trash bin icon next to it in Travis-CI).

I also added the Codecov application to GitHub (https://github.com/apps/codecov), but the only difference that I noticed between my test application and the results in this pull request, is that I have build checks for Codecov.

Example screenshot below

image

oliver-sanders:
Nice, can we turn off the comments for Codecov but still view it in the "checks" dialogue at the bottom. Codacy comments are driving me nuts at the moment.

kinow:
@oliver-sanders sure thing.

Here's an example where the coverage goes beyond a certain threshold, failing one of the checks in GitHub.

image

And here's another example where the coverage passes the minimum specified threshold.

image

In order to achieve this we need a `.codecov.yml` file, with some configuration settings - which I think should be OK. Will update this branch/pull request to include that file... not sure if Codecov will remove its last comment, but hopefully we shouldn't get any more comments from now on 👍

.travis.yml
```diff
@@ -3,11 +3,48 @@
...
+    minutes=0
+    limit=40
+    while kill -0 $! >/dev/null 2>&1; do
+      echo -n -e " \b" # never leave evidences!
+      if [ $minutes == $limit ]; then
+        break;
+      fi
+      minutes=$((minutes+1))
+      sleep 60
+    done
```
matthewrmshin: Can we just use `timeout 40m coverage ...` here?

kinow: What sorcery is it?! I never heard about `timeout`. First I googled for Travis and timeout, as I thought it would be something from Travis... but then realized it could be actually one of the many great GNU core utils that I am still not familiar with! http://www.gnu.org/software/coreutils/manual/coreutils.html#timeout-invocation #TodayILearned

Will finish reading the man page, and play with timeout in a bit. If it's not too late - and it works - will update it to use `timeout` instead in the next minutes. Otherwise it should be ready tomorrow :-) Thanks @matthewrmshin !!!

kinow: Really simple, didn't spend nearly as close as I thought I would need learning the basics of this tool. Will definitely remember to use it when I have a similar use case. Thanks a lot @matthewrmshin 

Left a [note](https://github.com/travis-ci/travis-ci/issues/4190#issuecomment-426972986) in the Travis-CI issue where I found the previous solution. Hopefully that will help others that like me didn't know `timeout`.

Pushed a new commit with the fix, and also rebased the branch.

kinow: Ooohh, true. I remember when I initially had the problem it was caused by Travis aborting the build. Not a problem, I ended up learning about timeout which will be handy in the future for sure. Reverting it :-)

oliver-sanders: Just background a process which calls `echo -n -e " \b"` every now an again?

kinow: Yup. Travis aborts the build if there's no output for more than 10 minutes. 

Coverage added more time to a test that took a while already, resulting in no output for a little over 10 minutes.

Travis provides `travis_wait` which works like `timeout`, preventing the job of being aborted.

However, `travis_wait` does not output anything. Even though the tests emit some useful info - especially when it fails! - so another developer came up with this poor-man's `timeout` variant in a github issue in travisci repo (link in the comment).

And I shamelessly copied the code here, changing only the timeout to 40 minutes (build took between 22-30 minutes IIRC).

.codecov.yml
```diff
@@ -0,0 +1,64 @@
...
+      default:
+        # commits below this threshold will be marked as failed
+        target: '75%'
+        # how much we allow the coverage to drop
+        threshold: '5%'
+    patch:
+      default:
+        # basic
+        target: '75%'
+        threshold: '5%'
```
oliver-sanders: I guess we will probably want to tweek this number as we go.

kinow: :+1 I guessed a number, lower than what we have right now. But I thought it would be easier to start with the threshold low, and then someone else with more experience with the code to fine-tune it.

kinow:
Thanks! 👏

PR #94, opened 2018-10-01
Use an enum for the timezone format mode

kinow:
Probably a very low priority pull request.

While writing the tests for `timezone.py`, and more specifically `get_local_time_zone_format`, I realized the parameters used in the function could be replaced by an enum.

The rationale for that, comes from the following example.

```python
Python 2.7.15rc1 (default, Apr 15 2018, 21:51:34) 
[GCC 7.3.0] on linux2
>>> import timezone
>>> timezone.get_local_time_zone_format()
'+1300'
>>> timezone.get_local_time_zone_format(False, False)
'+1300'
>>> timezone.get_local_time_zone_format(True, False)
'+13:00'
>>> timezone.get_local_time_zone_format(False, True)
'+13'
>>> timezone.get_local_time_zone_format(True, True)
'+13'
>>> 
```

The last case shows that we can specify `extended_mode=True` and `reduced_mode=True`, which is not possible. In reality, the code will ignore the `extended_mode` and simply use `reduced_mode`.

Even though I think this function is probably more for internal use, I think making sure we have an `Enum` or that we refactor it to use another approach, would prevent the introduction of accidental bugs, as well as make it easier for contributors to understand internal parts of the API.

As mentioned in the first paragraph, it should be covered by the tests written in another branch, pending #93 to be put in a pull request. Will have to update the tests in case if that's merged, if that's OK. Happy to change the approach here, or if others prefer other alternatives, or a simple won't do it (no hard feelings :grin: )

Cheers
Bruno


kinow:
Added WIP to the title as it looks like even though it was backported to Python 2.x, Enum's are still not automatically available. So I think if we decide to do it, it would have to wait until we port to Python 3 :disappointed:

matthewrmshin:
(OK. Only merge when Python 3 port is under way.)

oliver-sanders:
Just take out the `Enum` dependency, it doesn't really do anything much.

```diff
- from enum import Enum

- class TimeZoneFormatMode(Enum):
+ class TimeZoneFormatMode(object):
    normal = 0
    reduced = 1
    extended = 2
```

Works the same and is Python2/3 compatible.

kinow:
Just logged off otherwise would give it a try right now. Will update the branch romorrow morning. Thanks @oliver-sanders!!!

kinow:
Postponed touching this one to fix the code after we had the tests in (pull request #95 )

kinow:
Updated the code as @oliver-sanders suggested (thanks again btw). But expecting some conflicts with #95

oliver-sanders:
@kinow Once we have the next `isodatetime` release and are looking at updating the version bundled with Cylc we will need a corresponding PR in Cylc (due to the removal of the `extended_mode`, `reduced_mode` kwargs).

PR #96, opened 2018-10-01
Prevent cache len going over MAX_CACHE_SIZE

kinow:
Hi,

In `util.py`  the decorator `cache_results` stores the results for given inputs. It has a constant `MAX_CACHE_SIZE`, pre-set at `100000`.

The docs for that decorator say

>A maximum of MAX_CACHE_SIZE arg-value pairs are stored

But after writing the tests I realized it is actually a maximum of MAX_CACHE_SIZE + 1. When the cache is full, and you call the function one more time, the cache won't be evicted.

Instead, the new value is added, and then only later, after another call is made, then the eviction happens... but at that point the cache will go back to MAX_CACHE_SIZE +1.

This pull request changes that behaviour. Tests should pass in this branch as there was no tests confirming the behaviour of the cache (though the file was almost 100% covered from uses in other parts of the project). But I have a new test waiting for this ticket stashed over here https://github.com/kinow/isodatetime/commit/dc27edb13a83b98ca1d0f9267eb3d568316c38af

But if others prefer to leave this as-is, I would be glad to just update the test to make it pass with MAX_CACHE_SIZE + 1 instead (and maybe update the docs as well?).

Cheers
Bruno


matthewrmshin:
@oliver-sanders please sanity check this one.

PR #99, opened 2018-10-02
Add unit tests for util.cache_results

kinow:
Simple tests for the cache class in `isodatetime.util.cache_results`. With #96 merged, these tests should guarantee verifying the cache features after the port to Python3.

Thanks
Bruno


matthewrmshin:
Travis CI failure caused by style issue?

kinow:
Ops! Fixed, thanks for spotting it @matthewrmshin !

PR #101, opened 2018-10-04
Invert order of checks when time ticks

kinow:
Hi @oliver-sanders 

I was browsing issues when I found #80 reported via a Cylc issue. I was curious to see in which part of the code it was happening. But then after debugging it once, it occurred to me that a possible solution would be to invert the order of the checks (i.e. the `if/else`'s) when the time is "ticked" and corrected when adding truncated `TimePoint`s.

But... even though I wrote a quick test and the possible fix, I am still not confident that that's the solution. Would you mind have a look at this one whenever you have time, please?

I am assuming that different than

>2000-01-01T00:00:00Z + 03-30T00:00:00Z = 2000-03-02T00:00:00Z

the expected result was actually

>2000-01-01T00:00:00Z + 03-30T00:00:00Z = 2000-03-30T00:00:00Z

If this assumption is wrong, then you probably don't even have to spend time looking at the code/test :grimacing: 

Cheers
Bruno

ps: in the Cylc ticket you mentioned other cases where the result was strange... I haven't tested those scenarios... maybe it could fix that too, or there could be other parts that need some fix...

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -1158,58 +1158,71 @@ class TimePoint(object):
                       minute_of_hour=None, second_of_minute=None):
         """Combine this TimePoint with truncated time properties."""
         new = self.copy()
+        target = TimePoint(year=None, month_of_year=month_of_year,
+                 week_of_year=week_of_year, day_of_month=day_of_month,
+                 day_of_week=day_of_week, hour_of_day=hour_of_day,
+                 minute_of_hour=minute_of_hour,
+                 second_of_minute=second_of_minute, truncated=True)
 
+        # actions that do not tick the clock
         if year_of_century is not None:
             new.to_calendar_date()
             new_year_of_century = new.year % 100
             while new_year_of_century != year_of_century:
                 new.year += 1
                 new_year_of_century = new.year % 100
+            target.year = new.year
         if year_of_decade is not None:
             new.to_calendar_date()
             new_year_of_decade = new.year % 10
             while new_year_of_decade != year_of_decade:
                 new.year += 1
                 new_year_of_decade = new.year % 10
+            target.year = new.year
+        if ((hour_of_day is not None or minute_of_hour is not None) and
+                second_of_minute is None):
+            second_of_minute = 0
+            target.second_of_minute = 0
+        if hour_of_day is not None and minute_of_hour is None:
+            minute_of_hour = 0
+            target.minute_of_hour = 0
+        if second_of_minute is not None or minute_of_hour is not None:
+            new.to_hour_minute_second()
+            target.to_hour_minute_second()
+
+        # any change here will need to tick and check
         if month_of_year is not None:
             new.to_calendar_date()
             while new.month_of_year != month_of_year:
                 new.month_of_year += 1
-                new.tick_over()
+                new.tick_over(target)
         if week_of_year is not None:
             new.to_week_date()
             while new.week_of_year != week_of_year:
                 new.week_of_year += 1
-                new.tick_over()
+                new.tick_over(target)
         if day_of_month is not None:
             new.to_calendar_date()
             while new.day_of_month != day_of_month:
                 new.day_of_month += 1
-                new.tick_over()
+                new.tick_over(target)
         if day_of_week is not None:
             new.to_week_date()
             while new.day_of_week != day_of_week:
                 new.day_of_week += 1
-                new.tick_over()
+                new.tick_over(target)
         if hour_of_day is not None:
             while new.hour_of_day != hour_of_day:
                 new.hour_of_day += 1.0
-                new.tick_over()
+                new.tick_over(target)
         if minute_of_hour is not None:
             while new.minute_of_hour != minute_of_hour:
                 new.minute_of_hour += 1.0
-                new.tick_over()
+                new.tick_over(target)
         if second_of_minute is not None:
             while new.second_of_minute != second_of_minute:
                 new.second_of_minute += 1.0
-                new.tick_over()
-        if second_of_minute is not None or minute_of_hour is not None:
-            new.to_hour_minute_second()
-        if ((hour_of_day is not None or minute_of_hour is not None) and
-                second_of_minute is None):
-            second_of_minute = 0
-        if hour_of_day is not None and minute_of_hour is None:
-            minute_of_hour = 0
+                new.tick_over(target)
         return new
 
     def __add__(self, other, no_copy=False):
@@ -1421,7 +1434,7 @@ class TimePoint(object):
         if was_week_date:
             self.to_week_date()
 
-    def tick_over(self):
+    def tick_over(self, target=None):
         """Correct all the units going from smallest to largest."""
         if (self.hour_of_day is not None and
                 self.minute_of_hour is not None):
@@ -1461,7 +1474,7 @@ class TimePoint(object):
             self.week_of_year += num_weeks
             self.day_of_week = days + 1
         if self.day_of_month is not None:
-            self._tick_over_day_of_month()
+            self._tick_over_day_of_month(target)
         if self.day_of_year is not None:
             while self.day_of_year < 1:
                 days_in_last_year = get_days_in_year(self.year - 1)
@@ -1488,7 +1501,13 @@ class TimePoint(object):
                 self.month_of_year -= CALENDAR.MONTHS_IN_YEAR
                 self.year += 1
 
-    def _tick_over_day_of_month(self):
+    def _tick_over_day_of_month(self, target=None):
+        """
+        Tick over day of month, using target as limit if provided.
+        :param target: target time point
+        :type target: TimePoint
+        :return:
+        """
         if self.day_of_month < 1:
             num_days = 2
             for month, day in iter_months_days(
@@ -1512,7 +1531,9 @@ class TimePoint(object):
                         if num_days == self.day_of_month:
                             break
                 self.year = start_year
-                self.month_of_year = month
+                if (target is None or
+                    self.month_of_year != target.month_of_year):
+                    self.month_of_year = month
                 self.day_of_month = day
         else:
             month_index = (self.month_of_year - 1) % CALENDAR.MONTHS_IN_YEAR
@@ -1528,7 +1549,9 @@ class TimePoint(object):
                         day_of_month=1):
                     num_days += 1
                     if num_days == self.day_of_month:
-                        self.month_of_year = month
+                        if (target is None or
+                                self.month_of_year != target.month_of_year):
+                            self.month_of_year = month
                         self.day_of_month = day
                         break
                 else:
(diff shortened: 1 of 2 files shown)


matthewrmshin:
@benfitzpatrick do you mind commenting on this one?

isodatetime/data.py
```diff
@@ -1162,62 +1162,62 @@ def add_truncated(self, year_of_century=None, year_of_decade=None,
                       minute_of_hour=None, second_of_minute=None):
         """Combine this TimePoint with truncated time properties."""
         new = self.copy()
-        if hour_of_day is not None and minute_of_hour is None:
```
benfitzpatrick: I definitely agree there is a bug in the current implementation.

The main idea is that the timepoint is incremented until the data matches the truncated data... so the properties specified in the truncated timepoint should equal the ones in the final timepoint (assuming they are self-consistent).

isodatetime/tests.py
```diff
@@ -1204,6 +1204,32 @@ def test_timepoint_time_zone(self):
...
+        t3 = t1 + t2
+        self.assertEqual(2000, t3.year)
+        self.assertEqual(3, t3.month_of_year)
+        self.assertEqual(30, t3.day_of_month)
+        self.assertEqual(0, t3.hour_of_day)
+        self.assertEqual(0, t3.minute_of_hour)
+        self.assertEqual(0, t3.second_of_minute)
+        self.assertEqual(data.Duration(hours=0, minutes=0), t3.get_time_zone())
+        self.assertEqual("2000-03-30T00:00:00Z", str(t3))
+
```
benfitzpatrick: I think maybe a few more tests would be good (sorry there were none before) - there are quite a few edge cases. 

I tried:

```diff
diff --git a/isodatetime/tests.py b/isodatetime/tests.py
index 54f9329..0fa22b7 100644
--- a/isodatetime/tests.py
+++ b/isodatetime/tests.py

@@ -1230,6 +1234,45 @@ class TestSuite(unittest.TestCase):
         self.assertEqual(data.Duration(hours=0, minutes=0), t3.get_time_zone())
         self.assertEqual("2000-03-30T00:00:00Z", str(t3))
 
+    def test_timepoint_adding_truncated_timepoint_new_years_eve(self):
+        """Test rolling over the year."""
+        t1 = data.TimePoint(year=2000, month_of_year=12, day_of_month=31,
+                            hour_of_day=0, minute_of_hour=0,
+                            second_of_minute=0, time_zone_hour=0,
+                            time_zone_minute=0)
+        t2 = data.TimePoint(month_of_year=1, day_of_month=1,
+                            hour_of_day=0, minute_of_hour=0,
+                            second_of_minute=0, time_zone_hour=0,
+                            time_zone_minute=0, truncated=True)
+        t3 = t1 + t2
+        self.assertEqual(2001, t3.year)
+        self.assertEqual(1, t3.month_of_year)
+        self.assertEqual(1, t3.day_of_month)
+        self.assertEqual(0, t3.hour_of_day)
+        self.assertEqual(0, t3.minute_of_hour)
+        self.assertEqual(0, t3.second_of_minute)
+        self.assertEqual(data.Duration(hours=0, minutes=0), t3.get_time_zone())
+        self.assertEqual("2001-01-01T00:00:00Z", str(t3))
+
+    def test_timepoint_adding_truncated_timepoint_new_years_eve_hour(self):
+        """Test rolling over the year and hour."""
+        t1 = data.TimePoint(year=2000, month_of_year=12, day_of_month=31,
+                            hour_of_day=23, minute_of_hour=0,
+                            second_of_minute=0, time_zone_hour=0,
+                            time_zone_minute=0)
+        t2 = data.TimePoint(hour_of_day=0, minute_of_hour=20,
+                            second_of_minute=0, time_zone_hour=0,
+                            time_zone_minute=0, truncated=True)
+        t3 = t1 + t2
+        self.assertEqual(2001, t3.year)
+        self.assertEqual(1, t3.month_of_year)
+        self.assertEqual(1, t3.day_of_month)
+        self.assertEqual(0, t3.hour_of_day)
+        self.assertEqual(20, t3.minute_of_hour)
+        self.assertEqual(0, t3.second_of_minute)
+        self.assertEqual(data.Duration(hours=0, minutes=0), t3.get_time_zone())
+        self.assertEqual("2001-01-01T00:20:00Z", str(t3))
+
     def test_timepoint_dumper(self):
         """Test the dumping of TimePoint instances."""
         parser = parsers.TimePointParser(allow_truncated=True,
```

I get:

```
  File "/opt/iso/isodatetime/tests.py", line 1245, in test_timepoint_adding_truncated_timepoint_new_years_eve
    self.assertEqual(1, t3.month_of_year)
  File "/opt/iso/isodatetime/tests.py", line 1028, in assertEqual
    super(TestSuite, self).assertEqual(test, control, superinfo)
AssertionError: 1 != 2
```

I think this is because the month is incremented first, which gives the 31st of January, then the day is incremented to get to the 1st of the month.

isodatetime/tests.py
```diff
@@ -1204,6 +1204,32 @@ def test_timepoint_time_zone(self):
                             duration, data.Duration(days=0),
                             i_test_date_str + " - " + j_test_date_str)
 
+    def test_timepoint_adding_truncated_timepoint(self):
+        """See issue #80. When adding a truncated time, the result timepoint
+        must not be inconsistent, e.g.
+
+        2000-01-01T00:00:00Z + 03-30T00:00:00Z = 2000-03-02T00:00:00Z
```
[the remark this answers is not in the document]
kinow: It isn't. I included this as an example of an error, with the expected correct behaviour in the line that follows (bug: was 2000-03-01T00:00:00Z, but expected 2000-03-30T00:00:00Z). But now that you mentioned it, I think this can be confusing, and not really helpful to first see a bad example, so replacing it by just the expected value :+1:

kinow:
I think I now understand a bit better how the previous code worked @benfitzpatrick , so reverted my change, leaving only the tests. Will need some more time, but I think it might be possible to find a simpler fix.

This fix should make all the three tests in this current pull request to work. Once that's done, I will add a few more tests, and if no other bugs appear, then it will be ready for another review :+1: 

Thanks!
Bruno

isodatetime/tests.py
kinow: Added the tests to the pull request. Travis-CI won't be happy, but once a proper fix is pushed here, it should build again fine.

kinow:
I think I know one way to fix it, but it would take a while to implement. It would change the order of checks to go from the largest to smallest (easy to implement), but then in `data.py`'s `TimePoint.tick_over()` I would have to change the logic to also match this behaviour. Yesterday when flying from Wellington back to Auckland I did a few tests on a piece of paper, and it apparently works for both the current bug, and also for the new tests added by @benfitzpatrick 

Will update the pull request later. First will focus on Python 3 and packaging :+1: 

Cheers
Bruno

benfitzpatrick:
Thanks very much!

lhuggett:
This is also an issue with ordinal date when going from leap-year to non-leap-year (and vice-versa), and week/day of week, since the ordinal date should be found in the destination year, not the initial year. The issue should be fixed by this change, since year needs to be ticked over before day-of-year or week-of-year. I suggest adding a couple of tests for these. Something like:
```
def get_truncated_timepoint_add_tests():
    """Yield tests for adding a truncated timepoint to a timepoint."""
    return [
        ("2010-06-15T12:29:38Z",
         "-034",
         "2011-02-03T12:29:38Z"),
        ("2010-06-15T12:29:38Z",
         "08354",
         "2108-12-19T12:29:38Z"),
    ]
```
in the timepoints to test section, and then an extra function
```
    def test_truncated_timepoint_add(self):
        """Test adding a truncated time point to a time point."""
        parser = parsers.TimePointParser(
            allow_truncated=True)
        for test_props1, test_props2, ctrl_string in (
                get_truncated_timepoint_add_tests()):
            point1 = parser.parse(test_props1)
            point2 = parser.parse(test_props2)
            test_point = (point1 + point2).to_calendar_date()
            test_string = str(test_point)
            self.assertEqual(test_string, ctrl_string,
                             "%s + %s" % (point1, point2))
```
in the TestSuite class.

kinow:
@benfitzpatrick spent a couple of hours this morning scratching my head and running tests... trying to find a solution with as little changes as possible...

Found one way through a *target* object, used to prevent the timepoint + truncated timepoint operation of passing over the limits of what's defined in the truncated timepoint. Implemented it with an optional parameter, so that other parts of the code are not changed.

All tests passing, but wonder if you have any more tests that you can think of, that would be useful for this pull request?

kinow:
@lhuggett I had a look at the tests, and the `-034` I believe stands for `day_of_year`. At least that's where that value was put after the parser processed it.

Debugging further ahead, I noticed that actually the `add_truncated` function could work with this new test and with the changes in this pull request in theory, as you suggested.

However, the current `add_truncated` function doesn't appear to support a `day_of_year` yet: https://github.com/metomi/isodatetime/blob/c310af564f717589ed0f634a4665f116fff8be56/isodatetime/data.py#L1155

I think it would be simpler to leave it for another issue after this one. What do you think?

benfitzpatrick:
I think it is tricky doing it this way around. I managed to find a test that won't work - if you leave a 'time-bomb' the date can get messed up:

```python
    def test_timepoint_adding_truncated_timepoint_jan_to_feb(self):
        """Test rolling over the month and day of month."""
        t1 = data.TimePoint(year=2000, month_of_year=1, day_of_month=31,
                            hour_of_day=23, minute_of_hour=0,
                            second_of_minute=0, time_zone_hour=0,
                            time_zone_minute=0)
        t2 = data.TimePoint(month_of_year=2, day_of_month=28,
                            hour_of_day=0, minute_of_hour=20,
                            second_of_minute=0, time_zone_hour=0,
                            time_zone_minute=0, truncated=True)
        t3 = t1 + t2
        self.assertEqual(2000, t3.year)
        self.assertEqual(2, t3.month_of_year)
        self.assertEqual(28, t3.day_of_month)
        self.assertEqual(0, t3.hour_of_day)
        self.assertEqual(20, t3.minute_of_hour)
        self.assertEqual(0, t3.second_of_minute)
        self.assertEqual(data.Duration(hours=0, minutes=0), t3.get_time_zone())
        self.assertEqual("2000-02-28T00:20:00Z", str(t3))
```
fails with '28 != 29' in the day_of_month ☹️

kinow:
@benfitzpatrick thanks for the quick feedback. I found where to fix it, and even made the test pass... but then realized the amount of extra work to keep adding this `if not target ...`, and the complexity of methods (which are already complex enough) would increase a lot.

I think I will discard this approach of inverting order of checks + target, go back to the whiteboard and see what other options we would have. :+1:

kinow:
Will close this one as there are conflicts, and I never managed to solve the issues in the tests.

Some other day will review the ticket and try again. I will just add a note to the ticket about the interesting comments and possible cases for unit tests that are in this pull request, so that either myself or any other developer can have some background when working on this issue.

Issue #102, opened 2018-10-05
AttributeError when printing a TimePoint object


kinow:
When a `TimePoint` is copied, the `truncated_dump_format` is ignored. When dumping the copied `TimePoint`, an `AttributeError` will be raised and the `self.truncated_dump_format` will fail (not declared in the constructor, never initialized/created).

To reproduce it, you would need to first create a truncated `TimePoint`, copy it, and then try to dump it. Here's a sample Python code to reproduce it.

```python
from isodatetime import data
t1 = data.TimePoint(year=2000, truncated=True, truncated_dump_format='YYYY')
print(t1)
t2 = t1.copy()
print(t2) # error here
```

The code will print `0000` for the first time point, but then will emit the following error when trying to print the second time point:

```bash
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "isodatetime/data.py", line 1539, in __str__
    if self.truncated_dump_format and not override_custom_dump_format:
AttributeError: 'TimePoint' object has no attribute 'truncated_dump_format'
```

I have a simple fix for that and will submit it as a separate pull request.

The example here looks contrived, but I had this error while experimenting with isodatetime and `TimeRecurrence` with leap years (after I saw a ticket in Cylc about it). You can have the same problem if you calculate the next time point with the following example:

```
t1 = data.TimePoint(year=2000, month_of_year=1, day_of_month=1, truncated=True, truncated_dump_format='YYYY')
r = data.TimeRecurrence(repetitions=2, start_point=t1, end_point=None, duration=data.Duration(days=100))
print(r.get_next(t1))
```

matthewrmshin:
Fixed by #103.

PR #103, opened 2018-10-05
Fix TimePoint dumper behaviour after the TimePoint object has been copied

kinow:
A possible fix for #102 

Simply includes the `truncated_dump_format` to the `DATA_ATTRIBUTES` dict. A test that fails without the change, and works after the change in `data.py`.

Cheers
Bruno

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -717,10 +717,10 @@ class TimePoint(object):
         "day_of_year", "day_of_month", "day_of_week",
         "week_of_year", "hour_of_day", "minute_of_hour",
         "second_of_minute", "truncated", "truncated_property",
-        "dump_format", "time_zone"
+        "truncated_dump_format", "dump_format", "time_zone"
     ]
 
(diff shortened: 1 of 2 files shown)


isodatetime/data.py
```diff
@@ -717,7 +717,7 @@ class TimePoint(object):
         "day_of_year", "day_of_month", "day_of_week",
         "week_of_year", "hour_of_day", "minute_of_hour",
         "second_of_minute", "truncated", "truncated_property",
-        "dump_format", "time_zone"
+        "truncated_dump_format", "dump_format", "time_zone"
     ]
 
     __slots__ = DATA_ATTRIBUTES + ["truncated_dump_format"]
```
oliver-sanders: If we are adding `truncated_dump_format` to `DATA_ATTRIBUTES` we no longer need to manually add it in `__slots__` (we will now get a duplicate item) so we can just do `__slots__ = DATA_ATTRIBUTES`

kinow: Good catch! Just did the change locally, all tests passing, updating the pull request.

Thanks!

PR #104, opened 2018-10-12
Add tests for data.py

kinow:
Hi,

Found what appears to be a small bug, as a `__floordiv__` wouldn't return anything in a certain situation.

But this pull request includes not only that fix, but also tests that cover all the divisions that I found in the last file pending some coverage review, `data.py`.

After this I believe we should have enough coverage to start working on Python 3 🎉 🎉 🌮 

Cheers
Bruno


matthewrmshin:
@sadielbartholomew please sanity check.

PR #108, opened 2018-10-19
Refix ordinal date bug

lhuggett:
Re-fix of accidentally reverted change in #106 plus tests on that section of code.


oliver-sanders:
Re-fixed!

Issue #110, opened 2018-12-11
Replace type checker function in Python 3.7


kinow:
When using Python 3.7, the `_type_checker` can go away and be replaced by type hints plus some checks for `None`.

Only issue with that is that the exception raised would change from `BadInputError` we would get `TypeError` I believe. But this way we wouldn't be duplicating something that can be done in the language (and actually, once type hinting was added to the project, parts of that function would never be used/covered in tests).

kinow:
The same can be said about other type casting functions such as `_int_caster`. These may go too, unless they are indeed parsing values that were stored as strings, or parsed from a string.

oliver-sanders:
@kinow is this closed by #123 ?

kinow:
Not exactly, I think that was a failed attempt. Later realized it would still be safer to leave the runtime checks. Probably the easiest is to close it for now.

PR #111, opened 2018-12-11
Update to Python 3.7+

kinow:
This pull request:

- updates Travis-CI to run `pycodestyle` before tests, and include Python 3.7 in tests
- also fixes a logic problem in Travis-CI script that would always mark the build as success, even when there was an error
- replaces `cmp` usage by the new [operators](https://docs.python.org/3/library/operator.html)
- replaces `basestring` by `str`
- replaces `xrange` by `range`
- replaces `mock` by `unittest.mock`
- replaces `__nonzero__` by `__bool__`
- removes strings that had the prefix `u"..."` as strings are [unicode by default](https://docs.python.org/3/howto/unicode.html#the-string-type) now in Python 3
- Use the old division `//` where appropriate instead of `/` which would return a float instead of a rounded down integer as in Python 2
- Adds a few more improvements and fixes, especially around iterators/views/generators instead of pure data structures

Note: it would be worth besides the code review, to check the coverage reports to see if we have any code not covered by tests, that could lead to bugs in Python 3. From a quick inspection, I found only one suspect: https://codecov.io/gh/metomi/isodatetime/src/master/isodatetime/data.py#L951

And thanks a ton to @matthewrmshin for jumping in and fixing logic issues, shell script, and other Python issues from my initial try. I've edited the commits to remove old or hack-ish commits, leaving the ones that are pertinent to the issue.

Left multiple commits in case we have to edit or drop them, and so others can also review parts of the change if necessary.

Added #110 too, as we can work on it later if necessary (adding types is not a must-have for `isodatetime` for now IMO, but would be nice for later?). There is also some code in the unit tests that can be replaced or updated now that we have unittest+mock with Python3 (e.g. I think some `assertABC` methods for instance).

And finally, once this goes in, we would also have to update `setup.py`.

Thanks
Bruno


isodatetime/parser_spec.py
```diff
@@ -171,69 +171,69 @@
...
+    (r"Www", r"W(?P<week_of_year>[0-9][0-9])",
      "W%(week_of_year)02d", "week_of_year"),
-    ("D", r"(?P<day_of_week>\d)",
+    (r"D", r"(?P<day_of_week>[0-9])",
      "%(day_of_week)01d", "day_of_week"),
-    ("z", r"(?P<year_of_decade>\d)",
+    (r"z", r"(?P<year_of_decade>[0-9])",
      "%(year_of_decade)01d", "year_of_decade"),
-    ("^---", "(?P<truncated>---)",
+    (r"^---", "(?P<truncated>---)",
```
matthewrmshin: Should `r"..."` the second item here, for consistent style.

isodatetime/parsers.py
```diff
@@ -344,7 +344,7 @@ def _parse_from_custom_regex(self, regex, data_string, dump_format=None,
         if not result:
             raise StrptimeConversionError(source, data_string)
         info = result.groupdict()
-        for property_, value in info.items():
+        for property_, value in info.copy().items():
```
matthewrmshin: Probably my fault for introducing an inconsistent style here. We should either use `dict.copy().items()` or `list(dict.items())`.

isodatetime/tests.py
```diff
@@ -578,8 +580,9 @@ def get_timepointparser_tests(allow_only_basic=False,
                         for time_zone_expr, time_zone_info in time_zone_items:
                             tz_expr = combo_expr + time_zone_expr
                             tz_info = {}
-                            for key, value in (combo_info.items() +
-                                               time_zone_info.items()):
+                            for key, value in \
+                                list(combo_info.items()) + \
+                                    list(time_zone_info.items()):
```
matthewrmshin: Maybe use `chain` as well?

isodatetime/tests.py
```diff
@@ -602,8 +605,9 @@ def get_timepointparser_tests(allow_only_basic=False,
                 for time_zone_expr, time_zone_info in time_zone_items:
                     tz_expr = combo_expr + time_zone_expr
                     tz_info = {}
-                    for key, value in (combo_info.items() +
-                                       time_zone_info.items()):
+                    for key, value in \
+                        list(combo_info.items()) + \
+                            list(time_zone_info.items()):
```
matthewrmshin: Maybe use `chain` as well?

isodatetime/data.py
```diff
@@ -1315,25 +1332,137 @@ def get_props(self):
...
+            my_date = self.get_calendar_date()
+            other_date = other.get_calendar_date()
+        else:
+            my_date = self.get_ordinal_date()
+            other_date = other.get_ordinal_date()
+        my_datetime = list(my_date) + [self.get_second_of_day()]
+        other_datetime = list(other_date) + [other.get_second_of_day()]
+        return my_datetime > other_datetime
+
+    def __ge__(self, other: "TimePoint") -> bool:
```
matthewrmshin: I wonder if we can just do `return not self.__lt__(other)`?

.travis.yml
```diff
@@ -33,18 +35,19 @@ script:
     # So here we are using the approach documented in the issue
     # https://github.com/travis-ci/travis-ci/issues/4190.
     coverage run -m isodatetime.tests &
+    TEST_PID="$!"
     minutes=0
     limit=40
-    while kill -0 $! >/dev/null 2>&1; do
+    while kill -0 "${TEST_PID}" >'/dev/null' 2>&1; do
```
matthewrmshin: More natural to use `ps ...`?

isodatetime/data.py
kinow: Maybe for the `__gt__`. `functools.total_ordering` would simplify it, but at the cost of performance... so maybe we can implement

- `__eq__`
- `__lt__`

And return `not __eq__` for `__ne__`, `__eq__ and __lt__` for `__le__`, and so it goes? I read somewhere that `__ne__` would call `not __eq__` by default, and wasn't mandatory to implement it, but haven't tried it.

isodatetime/parser_spec.py
kinow: Good spot! Fixed!

isodatetime/parsers.py
kinow: I think `list(dict.items())` is safer to avoid problems in case we implement `__copy__`/`__deepcopy__`/... At least I think `list(dict.items())` is less likely to have any side-effects. Fixed!

isodatetime/tests.py
kinow: Really need to get used to `chain` and automatically recognize when it's necessary. Fixed! Thanks!

.travis.yml
kinow: +1. And if we are in no hurry, I can give it a try tomorrow, so that I practice some Shell, and then maybe you could review it, please? :grimacing:

isodatetime/tests.py
kinow: Fixed.

kinow:
@matthewrmshin 

>The only other remaining thing to do is to add Travis CI builds for Python 3.4 and 3.5.

I saw [your commit](https://github.com/matthewrmshin/isodatetime/commit/97189ddc7ce3c2d9663b1f760610c39bc98cb219) in your branch to add 3.4, 3.5, and saw that Travis-CI was [happy](https://travis-ci.org/matthewrmshin/isodatetime/builds/466892999) with this commit!

Then instead of just cherry-picking it, I decided to just have a look if it was worth adding [tox](https://tox.readthedocs.io/en/latest/). Coming from a language with normally a single tool as entry-point (normally Maven, but sometimes Gradle), I find it quite annoying to have so many tools in the build process, and can think of several reasons for not using tox.

However, it does one job quite well, which is create a temporary virtual environment with the Python versions listed, run a command (normally test/coverage/functional test/etc), **and install the application in that environment**.

The last part is important because it makes sure users will be able to not only build and run the tests in a specific Python version, but also to install with `pip install isodatetime`, `python setup.py install`, etc.

As an example of why it's useful, once I added that to a temporary `python3-tox` branch, it failed in all but Python 3.5: https://travis-ci.org/kinow/isodatetime/builds/467287417

Due to the _hack_ we have for version numbers in `isodatetime`. If I recall correctly, while packaging `isodatetime` the first time (and Cylc too I think), `setup.py` was not happy with the `0000.00.0` kind of version number.

I used an internal class of `setuptools` to shush the build process, but looks like it only works in 3.5 (and 2.7, at least in my environment) :cry: :disappointed: so now will need to tweak it a bit to work with all versions.

But I think it could be worth adding tox... WDYT? 

For users/developers, they could still run

- `python setup.py test`
- `python -m isodatetime.tests`
- or `tox`

Running `tox` will looks at `tox.ini` for the Python versions to build... and will try to find each version in your environment... mine has 3.6 (native) and 3.7 (anaconda), so it just skips 3.4 and 3.5.

```bash
kinow@kinow-VirtualBox:~/Development/python/workspace/isodatetime$ tox
GLOB sdist-make: /home/kinow/Development/python/workspace/isodatetime/setup.py
py34 create: /home/kinow/Development/python/workspace/isodatetime/.tox/py34
ERROR: InterpreterNotFound: python3.4
py35 create: /home/kinow/Development/python/workspace/isodatetime/.tox/py35
ERROR: InterpreterNotFound: python3.5
py36 inst-nodeps: /home/kinow/Development/python/workspace/isodatetime/.tox/.tmp/package/1/isodatetime-2018.11.0.zip
py36 installed: atomicwrites==1.2.1,attrs==18.2.0,coverage==4.5.2,isodatetime==2018.11.0,more-itertools==4.3.0,pluggy==0.8.0,py==1.7.0,pytest==4.0.1,pytest-cov==2.6.0,six==1.12.0
py36 run-test-pre: PYTHONHASHSEED='3594089636'
py36 runtests: commands[0] | /bin/bash ./.travis/test_coverage.sh
...
...
...
```

Or should we focus just on Python3 for now?

Thanks
B

matthewrmshin:
Probably worth concentrating on Python 3 for this PR. We can introduce *tox* as a separate PR?

kinow:
Sure thing @matthewrmshin . Will cherry-pick that commit from your branch, which worked on Travis for your fork. Then tomorrow should hopefully leave the alternative for `kill -0` with `ps` ready for review.

kinow:
Cherry-picked your commit (amended the title), and then added one more commit with my tentative fix for `kill -0 $PID` => `ps -p $PID`. I used the following script for testing.

```bash
#!/bin/bash

sleep 5 &
TEST_PID=$!
seconds=0
limit=2

while ps -p "${TEST_PID}" >/dev/null 2>&1; do
  echo "Still running"
  if ((seconds == limit)); then
    echo "Aborting!"
    exit 1
  fi
  seconds=$((seconds + 1))
  sleep 1
done

echo "BYE"
wait "${TEST_PID}"
```

Changing the value of `limit`, and doing `echo $?` to confirm the exit code. And also passed it through `shellcheck` to see if there was any issue left.

Ready for review! (I think) :tada:

isodatetime/data.py
oliver-sanders: We could just implement `__eq__` and `__lt__` as you suggest, there is no need to do `not __eq__` type stuff. If it's quick to do we should probably just go for it and implement the full array for the performance improvement though?

isodatetime/dumpers.py
```diff
@@ -226,7 +226,7 @@ def get_time_zone(self, time_zone_string):
             return None
         info = self._timepoint_parser.process_time_zone_info(info)
         if info.get('time_zone_utc'):
-            return (0, 0)
+            return 0, 0
```
oliver-sanders: Out of interest is this safety for Python3 or preference?

isodatetime/parser_spec.py
```diff
@@ -171,69 +171,69 @@
...
-    ("^--", "(?P<truncated>--)",
+    (r"^--", r"(?P<truncated>--)",
      "--", None),
-    ("^-", "(?P<truncated>-)",
+    (r"^-", r"(?P<truncated>-)",
      "-", None)
 ]
 _TIME_TRANSLATE_INFO = [
-    ("(?<=^hh)mm", r"(?P<minute_of_hour>\d\d)",
+    (r"(?<=^hh)mm", r"(?P<minute_of_hour>[0-9][0-9])",
```
oliver-sanders: Why move away from the `\d` shorthand?

matthewrmshin: It did not work, and gave us backslash hell. The `\d` version is also less safe because it matches Unicode numerics in other languages.

isodatetime/data.py
kinow: Resolving the conversation as we have dealt with this problem already :tada:

.travis.yml
kinow: Oh, this one we already tackled too. So marking as resolved to reduce the number of open conversations in the PR

isodatetime/parser_spec.py
kinow: Yup, we used the `\d` in a `regex.sub` I think, in a field that was not for regex, but for ordinary text. Problem was that we were using that method to craft the final pattern for a new regex.

Until some versions ago it was fine, but in Python 3.6 or 3.7 (can't remember which one) instead of a warning, it blew an error if I recall correctly. And the documentation of the Python function indicates that it won't be happy with characters with the `\` that are used for regex escape.

I fixed it with an ugly hack, but @matthewrmshin reviewing found a much saner solution of simply replacing the `\d` by `[0-9]`, which also increases the security for it ignores unicode numerics :clap:

isodatetime/dumpers.py
kinow: Oh, not necessary for Python 3. It was just among the warnings raised by my IDE. As I went through the warnings I tried to stick with the ones that were related to Py2 to Py3, but this one slipped in. I can remove these changes from the pull request if that's easier to review.

.travis.yml
```diff
@@ -5,25 +5,33 @@
 language: python
 
 python:
-  - 2.6
-  - 2.7
+  - 3.4
+  - 3.5
+  - 3.6
+  - 3.7-dev
```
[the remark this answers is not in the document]
kinow: I think this could work (https://docs.travis-ci.com/user/languages/python/#development-releases-support).

I will update the branch now and let Travis give it a try.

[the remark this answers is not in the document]
kinow: Had a look a few minutes after the build started, and alas looks like there's something weird with Python 3.8 in Travis. Two 3.8 builds immediately failed with

```bash
3.8-dev is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.8-dev.tar.bz2
$ curl -sSf -o python-3.8-dev.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 403 Forbidden
Unable to download 3.8-dev archive. The archive may not exist. Please consider a different version.
```

I remember seeing something similar with Python 3.7 in Travis. It was a problem with the container image that the build was running on, and updating (but it was to download an `apt` package). Upgrading to the latest fixed. But I think it might be easier to leave 3.8 for later.

Reverted the commit.

matthewrmshin:
Tests pass in my environment.

isodatetime/parser_spec.py
oliver-sanders: Dam it Python, I thought this was supposed to be getting easier! Taking a scan of the `re` docs they have introduced a `re.ASCII` flag which might be useful in other examples.

PR #113, opened 2019-01-09
Update classifiers and python-requires after Python 3

kinow:
Follow-up after #111, the `python-requires` definitely needs `>=3.4`. I've added the `< 3.8` as we were not able to test that in Travis-CI (yet!).

Also tried to update the classifiers accordingly... feel free to suggest different/other classifiers. A complete list can be found here https://pypi.org/pypi?%3Aaction=list_classifiers (good practice to keep them in order as they appear - I think I followed that rule as best as I could).

Cheers
Bruno

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ setup(
     url="https://github.com/metomi/isodatetime",
     packages=['isodatetime'],
     long_description=read('README.md'),
+    long_description_content_type="text/markdown",
     platforms='any',
     install_requires=[],
     python_requires='>=3.4, <3.8',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -51,15 +51,19 @@ def read(fname):
     long_description=read('README.md'),
```
matthewrmshin: Please also add:

```
long_description_content_type="text/markdown",
```

PyPI should then render our `README.md` correctly.

kinow: Good idea! Done! Travis will be busy for a while now.

matthewrmshin:
One review is good enough for this one.

PR #115, opened 2019-01-10
Change validation order

kinow:
Some time ago while reviewing a pull request, I noticed that `self.copy()`, which doesn't alter the current object's state, but rather just creates a new object.

Creating the object is a considerably expensive operation, if you compare against an `if` statement to validate something. So I thought changing the order in that case would make sense.

Then I thought it would be useful to have something to automatically detect it, and wrote https://github.com/kinow/pylint-validation-order, which is pretty much very ugly code, using `pylint` plug-in API, and `astroid` to parse Python AST, and identify when a validation `if` happens before a statement that has nothing to do with the validation. Obviously, it misses several cases, but still gives a somewhat useful output.

This pull request includes all occurrences of validation logic found with that tool, that I believe can be done in another order, to fail before calling operations that could create unnecessary objects.

Cheers
Bruno

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -496,13 +496,13 @@ class Duration(object):
 
     def __mul__(self, other):
         # TODO: support float multiplication?
-        new = self.copy()
         if not isinstance(other, int):
             raise TypeError(
                 "Invalid type for multiplication: " +
                 "'%s' should be integer." %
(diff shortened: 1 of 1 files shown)


oliver-sanders:
> and wrote https://github.com/kinow/pylint-validation-order

Nice!

Issue #47, opened 2014-06-04
Document basic usage


benfitzpatrick:
We should document some basic usage of the library in the README file.

matthewrmshin:
Close by #117.

PR #117, opened 2019-01-18
Readme: add installation instructions and examples

oliver-sanders:
Should suffice for now.


matthewrmshin:
(After this we'll need to do a happy new change + an update to CHANGES.md.)

oliver-sanders:
I've got `pytest` to run on all `.md` files so that the python examples in the `README.md` file get tested. Easy enough.

There are, however, some console examples which will go untested. As this is a problem we have elsewhere in Rose & Cylc I though it might be worth a look-in now that we are using an extendible testing framework.

So ... https://github.com/oliver-sanders/pytest_shelltest

matthewrmshin:
Is this sufficient to close #47?

kinow:
> Is this sufficient to close #47?

+1 from me. We can close that one now and add more to the documentation based on user/dev feedback later IMO

Issue #91, opened 2018-09-23
date_string referenced before assignment


kinow:
A user can programmatically set certain values to `None`, causing a runtime error in the API. Example test case here: https://github.com/kinow/isodatetime/commit/2e05d90aefa04d0e3b3fea320fd0bc93a2d3b506

Here's the same code from the commit above

```python
    def test_timepoint_dump_format(self):
        """Test the timepoint format dump when values are programmatically set to None"""
        t = data.TimePoint(year="1984")
        # commenting out month_of_year here is enough to make the test pass
        t.month_of_year = None
        t.day_of_year = None
        t.week_of_year = None
        self.assertEqual("1984-01-01T00:00:00Z", t.__str__())
```

I believe while a user may not have the need nor intention to set these values to `None`, it may still happen when using the API in another software.

So I _think_ maybe we could raise an error in the `_get_truncated_dump_format` if we were not able to create `date_string`, or find another way to prevent the invalid runtime state.

Bruno

matthewrmshin:
@kinow Is #118 not sufficient to fix this?

kinow:
Oh, it is!

PR #118, opened 2019-01-19
Check for date type before dumping values

kinow:
Tentative fix for #91. Not the best coming up with exception messages... probably needs some re-wording perhaps? Realized it was doing `if`, `if`, and `if`. But the `if`'s are exclusive, so it could be replaced by `if`/`elif`'s (I believe). Test included.

Cheers
Bruno

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -1720,10 +1720,14 @@ class TimePoint(object):
 
         if self.get_is_calendar_date():
             date_string = year_string + "-MM-DD"
-        if self.get_is_ordinal_date():
+        elif self.get_is_ordinal_date():
             date_string = year_string + "-DDD"
-        if self.get_is_week_date():
+        elif self.get_is_week_date():
             date_string = year_string + "-Www-D"
+        else:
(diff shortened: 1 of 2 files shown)


isodatetime/data.py
```diff
@@ -1720,10 +1720,14 @@ def _get_dump_format(self):
...
         if self.get_is_calendar_date():
             date_string = year_string + "-MM-DD"
-        if self.get_is_ordinal_date():
+        elif self.get_is_ordinal_date():
             date_string = year_string + "-DDD"
-        if self.get_is_week_date():
+        elif self.get_is_week_date():
             date_string = year_string + "-Www-D"
+        else:
+            raise RuntimeError("Inconsistent state before dumping. Must"
```
oliver-sanders: Perhaps:

> TimePoint has inconsistent state, points must conform to calendar, ordinal or week dates.

kinow: +1 So much better! Changing right now. Thanks!

kinow: Done, also rebased branch onto master. Travis should - hopefully - build it with no issues :crossed_fingers:

PR #119, opened 2019-01-20
Set integer type in get_time_zone before return

### LINKED ISSUE #97: Inconsistence in time zone values returned by dumpers.TimePointDumper.get_time_zone()
In `dumpers.TimePointDumper`, `get_time_zone` returns a time zone tuple for a given string. So `+01:00` will return a tuple with two strings `('01', '00')`.

While writing unit tests to cover some parts of the `TimePointDumper`, I realized my input values test matrix looked a bit weird. Here's the values used as input, with their equivalent returned values.

| input (string) | output (tuple or None) | notes |
|--------|---------|----------|
| `'+01:00'` | `('01', '00')` | `string`, `string` |
| `'+00:00'` | `('00', '00')` | `string`, `string` |
| `'-00:00'` | `(0, 0)` | `int`, `int` |
| `'-03:00'` | `(-3, 0)` | `negative int`, `int` |
| `'-03:30'` | `(-3, -30)` | `negative int`, `negative int` (maybe sign just hours?) |
| `'Z'` | `(0, 0)` | `int`, `int` |

I believe this is not breaking anything (at least I couldn't find anything in the existing issues), but it makes harder for users to use the values returned. I think consistency here would ease the use, and also reduce the chance of bugs.

Not a blocker for next release I think... and probably not a blocker for moving to Python 3, so just creating the issue for now :+1: 

Cheers
Bruno

kinow:
Fix #97 . This sets the time zone information's hour and minute tuple to `Tuple[int, int]`, instead of returning something like `Union[Tuple[str, str], Tuple[int, int]]`.

Updated test. Now `Z`, `+00:00`, and `-00:00` all return the same response: `(0, 0)`.

Cheers
Bruno


matthewrmshin:
@oliver-sanders please sanity check.

PR #122, opened 2019-03-08
Small Documentation fixes

wxtim:
I've removed a few things that have come to light in creating the new wrapper for rose-date:

1. Removed references to -c

2. Added `=` signs to `--offset` options since `--offset -PT2H` doesn't work with Argparse. Apparently Argparse doesn't like anything it doesn't think is a number being passed like this and ISO-Datetime-offsets aren't sufficiently number-like. :disappointed: 

3. Make argparse cope with intermixed args and options (at least at python 3.7).

### DIFF
--- a/isodatetime/main.py
+++ b/isodatetime/main.py
@@ -280,7 +280,7 @@ def main():
         ],
     ]:
         arg_parser.add_argument(*o_args, **o_kwargs)
-    if hasattr(arg_parse, 'parse_intermixed_args'):
+    if hasattr(arg_parser, 'parse_intermixed_args'):
         args = arg_parser.parse_intermixed_args()
     else:
         args = arg_parser.parse_args()
--- a/isodatetime/main.py
+++ b/isodatetime/main.py
@@ -280,7 +280,10 @@ def main():
         ],
     ]:
         arg_parser.add_argument(*o_args, **o_kwargs)
-    args = arg_parser.parse_args()
+    if hasattr(arg_parse, 'parse_intermixed_args'):
+        args = arg_parser.parse_intermixed_args()
+    else:
+        args = arg_parser.parse_args()
     if args.version_mode:
         print(__version__)
         return


isodatetime/main.py
```diff
@@ -225,7 +225,7 @@ def main():
             },
         ],
         [
-            ["--offset1", "--offset", "-s", "-1"],
+            ["--offset1=", "--offset=", "-s=", "-1="],
```
matthewrmshin: Do you actually need to do this?

isodatetime/main.py
```diff
@@ -285,7 +285,7 @@ def main():
         ],
     ]:
         arg_parser.add_argument(*o_args, **o_kwargs)
-    args = arg_parser.parse_args()
+    args = arg_parser.parse_intermixed_args()
```
matthewrmshin: Thought you were going to do `if hasattr(arg_parse, 'parse_intermixed_args'):` ...

isodatetime/main.py
wxtim: No!  I've undone it again!

isodatetime/main.py
wxtim: yes, I was. I've been making my own life very complex with editing the version embedded in rose vs editing isodatetime.git. Looking forward to packaging!

matthewrmshin:
@wxtim Tests are failing.

isodatetime/main.py
```diff
@@ -285,7 +285,10 @@ def main():
         ],
     ]:
         arg_parser.add_argument(*o_args, **o_kwargs)
-    args = arg_parser.parse_args()
+    if hasattr(arg_parse, 'parse_intermixed_args'):
```
oliver-sanders: Wazzthis about?

wxtim:
@matthewrmshin Tests passing

matthewrmshin:
1 review good enough for this.

PR #123, opened 2019-03-12
Use Python 3 typing instead of functions to check types

### LINKED ISSUE #110: Replace type checker function in Python 3.7

kinow:
Fixes #110

Instead of using the current type checking functions, in this pull request the parameter received type hints, specifying what possible types are acceptable through `Union[]`. Unit tests and functional tests in Travis passing.

Marked as draft as I have not tested with Cylc or in the Notebooks I had yet :+1:

### DIFF
--- a/isodatetime/data.py
+++ b/isodatetime/data.py
@@ -24,6 +24,8 @@ from . import timezone
 
 from functools import lru_cache
 
+from typing import Union
+
 
 class Calendar(object):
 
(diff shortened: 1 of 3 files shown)


kinow:
One downside is that for Python 3.4 we need to ask users to install `typing` module. I added that in Travis, and would possibly have to add that to `setup.py` when the user uses `3.4` or earlier (I have seen that done in some project I think, so should be doable).

isodatetime/data.py
```diff
@@ -146,175 +148,6 @@ def __str__(self):
         return format_string.format(*format_args)
 
 
-class TimeRecurrence(object):
```
kinow: Not deleted, only moved after `TimePoint` as it needs to be defined before it, so we can reference an argument to be typed as `TimePoint`.

isodatetime/data.py
```diff
@@ -741,74 +572,49 @@ class TimePoint(object):
...
-            (minute_of_hour, "minute_of_hour", None, int, float),
-            (minute_of_hour_decimal, "minute_of_hour_decimal", None, float),
-            (second_of_minute, "second_of_minute", None, int, float),
-            (second_of_minute_decimal, "second_of_minute_decimal", None,
-             float),
-            (time_zone_hour, "time_zone_hour", None, int),
-            (time_zone_minute, "time_zone_minute", None, int)
-        )
         self.dump_format = dump_format
-        self.expanded_year_digits = _int_caster(expanded_year_digits,
```
kinow: Oh, this will break later. Need to keep it as `int`.

kinow:
Closing as this wouldn't completely replace the runtime checks. Might have another go at it some other day :)

PR #124, opened 2019-03-15
use parse_intermixed_args in Python versions where this is available

wxtim:
[Whilst upgrading rose date to rely more heavily on isodatetime](https://github.com/metomi/rose/issues/2303) I added the following to increase the flexibility of isodatetime arguments and options.

[WIP] - Needs Tests


matthewrmshin:
Same change as #122.

PR #127, opened 2019-04-05
parsers: raise ValueError for invalid TimePoints

oliver-sanders:


matthewrmshin:
Tests unhappy. :sob:

kinow:
I tried changing the logic in the tick methods some time ago (#101) and found that for each scenario I added a fix, it actually broke the logic in some other scenarios :disappointed:

oliver-sanders:
Some tests broken due to timezone changes. It looks like this configuration does nothing:

```ini
 [pytest]
-env =
-    TZ=UTC
```

PR #131, opened 2019-06-25
Travis CI: use xenial dist

matthewrmshin:
Update distro on Travis CI, and include a Python 3.8-dev build.


matthewrmshin:
One review should be good enough.

PR #132, opened 2019-07-10
Metomi isodatetime

wxtim:
Convert isodatetime's  module name into metomi.isodatetime for consistency with rose and rosie.
- [x] Change the namespace
- [x] Check the tests


.travis.yml
```diff
@@ -8,8 +8,6 @@ language: python
 matrix:
   fast_finish: true
   include:
-    - python: 3.4
```
wxtim: I removed this because 3.4 is not supported by Pytest 5. I can probably work out how to add logic if you would prefer me to use a different pytest version, but it seems like a lot of work for an ageing Python Version.

matthewrmshin: The only reason is that the default version of Python on our RHEL7 machines is Python 3.4. But the reality is that we'll never use this library with that version of Python. So, I guess it is fair to drop 3.4.

matthewrmshin:
All looking good, and build in my environment.

metomi/isodatetime/tests/test_main.py
```diff
@@ -24,8 +24,8 @@
 from unittest.mock import patch
 
 
-import isodatetime
-import isodatetime.main
+import metomi.isodatetime as isodatetime
```
sadielbartholomew: A alias that does not clash with the ``isodatetime`` submodule name would be safer here, as there is the potential for future confusion or errors.

But, since other than ``isodatetime_main`` there is only one call to it, there seems little need to define an alias, just import what is required:
```suggestion
from metomi.isodatetime import __version__
```
then use that method directly on line 42.

wxtim: Yes.

wxtim:
@sadielbartholomew - think I've fixed that and await your approval. @matthewrmshin would you like me to put this up on pypi?

PR #134, opened 2019-07-12
Use old project name in setup.py?

kinow:
Hi,

A pull request with a question :grimacing: Just wondering if the project name was supposed to change with the new project structure/layout?

If so we will need to update the dependency in cylc-flow's `setup.py` and someone will need to reserve the name in PYPI too?

Cheers
Bruno

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,7 +63,7 @@ class bdist_rpm(bdist_rpm_original):
 
 
 setup(
-    name="metomi-isodatetime",
+    name="isodatetime",
     version='1!' + __version__,
     author="Met Office",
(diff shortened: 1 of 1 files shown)


kinow:
Ah! Went for a walk, and then on the way back realized that actually someone told me about `rose` being taken in PYPI, and that the project would have to be something like `metomi-rose`... I just didn't associate. That means everything under github.com/metomi will be published as `metomi-$PROJECT_NAME`?

oliver-sanders:
Yep, it's going to be:

* `metomi.isodatetime`
* `metomi.rose`
* `metomi.rosie`

PR #135, opened 2019-07-12
setup: fix version identification

oliver-sanders:
In the `setup.py` we have been determining the version via:

```python
from metomi.isodatetime import __version__
```

Which makes good sense, but, looking back at it again it seems a bit dodge, if we already have `isodatetime` installed will it not pick up that version then install the current code under that version?

Solution, import the `__init__.py` file by relative path.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,7 +37,7 @@ except FileNotFoundError:
 
 def get_version(module, path):
     """Return the __version__ attr from a module sourced by FS path."""
-    spec = spec_from_file_location(module, path)
+    spec = spec_from_file_location(module, str(path))
     module = module_from_spec(spec)
     spec.loader.exec_module(module)
     return module.__version__
--- a/setup.py
+++ b/setup.py
@@ -29,15 +29,12 @@ from setuptools import setup, find_namespace_packages
 import sys
 
 
-try:
-    os.chdir(os.path.dirname(__file__))
-except FileNotFoundError:
-    pass
+DIST_DIR = Path(__file__).resolve().parent
 
 
 def get_version(module, path):
     """Return the __version__ attr from a module sourced by FS path."""
-    spec = spec_from_file_location(module, str(path))
+    spec = spec_from_file_location(module, path)
     module = module_from_spec(spec)
     spec.loader.exec_module(module)
     return module.__version__
@@ -45,7 +42,7 @@ def get_version(module, path):
 
 __version__ = get_version(
     'metomi.isodatetime',
-    Path('metomi/isodatetime/__init__.py')
+    str(Path(DIST_DIR, 'metomi/isodatetime/__init__.py'))
 )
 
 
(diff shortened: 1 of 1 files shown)


matthewrmshin:
Tests break.

kinow:
Huh, one build in Travis failed with

```
coverage run setup.py test --pytest-args='--runslow' &
TEST_PID="$!"
minutes=0
limit=50
while ps -p "${TEST_PID}" >/dev/null 2>&1; do
  echo -n -e " \b" # never leave evidences!
  if ((minutes == limit)); then
    exit 1
  fi
  minutes=$((minutes + 1))
  sleep 60
done
wait "${TEST_PID}"
  File "setup.py", line 48, in <module>
    Path('metomi/isodatetime/__init__.py')
  File "setup.py", line 40, in get_version
    spec = spec_from_file_location(module, path)
  File "<frozen importlib._bootstrap_external>", line 581, in spec_from_file_location
AttributeError: 'PosixPath' object has no attribute 'endswith'
Coverage.py warning: Module ./isodatetime was never imported. (module-not-imported)
Coverage.py warning: No data was collected. (no-data-collected)
The command "export COVERAGE_PROCESS_START="${TRAVIS_BUILD_DIR}/.coveragerc"
```

So just kicked it to see it that does the trick.

setup.py
```diff
@@ -16,16 +16,37 @@
...
+    """Return the __version__ attr from a module sourced by FS path."""
+    spec = spec_from_file_location(module, path)
+    module = module_from_spec(spec)
+    spec.loader.exec_module(module)
+    return module.__version__
+
+
+__version__ = get_version(
+    'metomi.isodatetime',
+    Path('metomi/isodatetime/__init__.py')
```
kinow: Looks like it doesn't work for Python 3.5:

```
  File "setup.py", line 48, in <module>
    Path('metomi/isodatetime/__init__.py')
  File "setup.py", line 40, in get_version
    spec = spec_from_file_location(module, path)
  File "<frozen importlib._bootstrap_external>", line 581, in spec_from_file_location
AttributeError: 'PosixPath' object has no attribute 'endswith'
```

oliver-sanders: Hmmm? It works for me on Python3.4 (don't have 3.5 at hand):

```console
$ python3.4
Python 3.4.10 (default, Apr  8 2019, 02:17:36) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux
>>> from pathlib import Path
>>> Path('metomi/isodatetime/__init__.py')
PosixPath('metomi/isodatetime/__init__.py')
```

oliver-sanders: Ahhh, got it, the `spec_from_file_location` function only picked up Pathlib support at 3.6!

https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location

oliver-sanders: Have added Python3.5 support as it was simple to do.

setup.py
```diff
@@ -16,16 +16,37 @@
...
 # https://docs.pytest.org/en/2.8.7/goodpractices.html#manual-integration
 from setuptools.command.test import test as TestCommand
 from setuptools import setup, find_namespace_packages
+import sys
+
+
+try:
+    os.chdir(os.path.dirname(__file__))
+except FileNotFoundError:
+    pass
```
matthewrmshin: Currently, if the current working directory is not the same directory containing the library, the default behaviour for something like `python3 /path/to/metomi/isodatetime/setup.py bdist` is to build a distribution file under `$PWD/dist/`.

Does this change the behaviour? (May surprise some usage.)

matthewrmshin: @oliver-sanders This block is still changing the current working directory after your latest commit. Is this still necessary?

oliver-sanders: Forgot to commit the change which removed it.

PR #136, opened 2019-07-22
 updated for release

wxtim:
- [x] Added information to `CHANGES.md`
- [x] Changed version in `__init__.py` to 2.0.1
- [x] Checked and updated `CONTRIBUTING.md`

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,30 @@ for a full listing of issues for each release.
 
 --------------------------------------------------------------------------------
 
+## isodatetime 2.0.1 (Released 2019-07-23)
+
+This is the 13th release of isodatetime.
+
+This release requires Python 3.5 or above.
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -5,6 +5,30 @@ for a full listing of issues for each release.
 
 --------------------------------------------------------------------------------
 
+## isodatetime 2.0.1 (Released 2019-07-23)
```
kinow: No bug fixes, but some updated requirements... should it be 2.1.0 instead?

matthewrmshin: I am torn. You are probably right as we are changing the namespace etc. However, we don't really have real known users for this version of the library yet, so either way is probably OK.

kinow: Good point :+1:

PR #137, opened 2019-07-23
specify ISO8601:2004 standard

oliver-sanders:
ISO8601:2004 is now officially withdrawn so nowerdays ISO8601 really refers to ISO8601-1:2019.

Declare support for the ISO8601:2004 spec.

ISO8601-1:2019 is pretty much the same but we would need a code/spec review and some new tests to claim ISO8601-1:2019 support.


oliver-sanders:
https://github.com/metomi/isodatetime/issues/138

PR #144, opened 2019-09-20
Include LICENSE.txt in sdist and wheel

xhochy:
Ship license in all cases, this also removes the need to bundle the license with the conda recipe.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include LICENSE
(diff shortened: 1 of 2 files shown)


kinow:
Tested by running `python setup.py sdist` and `python setup.py bdist_wheel`. It is possible to confirm in the command output, as well as in the generated archives that the `LICENSE` file was included. Running without this, the `LICENSE` is omitted (at least in the `sdist` that I tested before).

Thanks @xhochy !!!

matthewrmshin:
@xhochy Many thanks!

PR #146, opened 2019-09-20
testing: mac os

oliver-sanders:
@kinow you said on Riot that you had hit an issue with Isodatetime and Mac OS which surprised me since isodatetime shouldn't interface with the OS directly.

I've worked on Mac OS builds with Travis before so thought I'd quickly add one here just to make sure. Feel free to reject if we don't think it's worth keeping.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,21 @@ matrix:
       env: RUN_COVERAGE=false TZ=UTC
     - python: 3.8-dev
       env: RUN_COVERAGE=false
+    - os: osx
+      osx_image: xcode9.4
+      # install python via homebrew, travis cannot do this atm
+      language: generic
+      addons:
+        homebrew:
+          update: true
+          packages: python3
+      # use a virtualenv for consistency with normal travis python installation
+      before_install:
+        - pip3 install virtualenv
+        - virtualenv -p python3 ~/venv
+        - source ~/venv/bin/activate
 
-before_install:
+install:
   # Upgrade pip and setuptools as some times the versioning scheme hack fails
   # in Travis-CI.
   - pip install --upgrade pip setuptools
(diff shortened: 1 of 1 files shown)


matthewrmshin:
Looks good. Waiting for test to complete.

matthewrmshin:
The MacOS stage says "no language set" instead of "python ..."?

.travis.yml
```diff
@@ -18,8 +18,21 @@ matrix:
       env: RUN_COVERAGE=false TZ=UTC
     - python: 3.8-dev
       env: RUN_COVERAGE=false
+    - os: osx
+      env: RUN_COVERAGE=false
+      # Install Python via Homebrew, Travis cannot install Python ATM.
+      language: generic
```
oliver-sanders: > The MacOS stage says "no language set" instead of "python ..."?

That would be because of this line.

**tldr;** *Travis can't install different python versions automatically you need to provision it yourself.*

**long overdue rant;** As is well known Python environments usually become a bit of a mess. On Mac OS the problem becomes slightly messier as Mac OS does not actually have a package manager, well not an official one anyway. Apple gives you system Python but if you want anything else there isn't an Apple provided solution to getting it. The on-official but incredibly popular Brew package manager might as well be dubbed "the" Mac OS package manager. Most people turn to Brew to install different Python versions, consequently the Brew formula is pretty good and actually surpasses the system installation in some ways. So Brew Python is what everyone should use but neither it nor Brew are actually installed by default leaving Travis in a slightly sticky position. Their solution for the moment is to do nothing which is great except that if you try to specify a Python version it just dies in a nonsensical way. The workaround to all this mess is to set the language to something other than Python and just install Brew Python yourself.

matthewrmshin:
The test has passed after being kicked...

Do we want to have a look at the hanging test problem before merge or after?

kinow:
>Do we want to have a look at the hanging test problem before merge or after?

Good question. Will leave it up to @oliver-sanders. If happy with the current approach, feel free to merge, or move the discussion to an issue, etc.

oliver-sanders:
I'm going to kick Travis a few times to see if it hangs again...

oliver-sanders:
OK, kicked a few times, can't replicate. Merging.

PR #153, opened 2019-11-06
Negative durations

datamel:
Fixed bug and close issue #152.
Added bug to check bug was fixed.

### DIFF
--- a/metomi/isodatetime/main.py
+++ b/metomi/isodatetime/main.py
@@ -48,9 +48,10 @@ SYNOPSIS
     isodatetime 19700101T000000Z 20380119T031407Z
 
     # 3.  Print ISO8601 duration as total amount of a unit
-    # 3.1 Into the total number of hours (H), minutes (M) or seconds (S)
+    #     Into the total number of hours (H), minutes (M) or seconds (S)
     #     it represents.
-    
+    isodatetime --as-total=s PT1H
+
     # 4.  Print a number of time points in a ISO8601 recurrence
     # 4.1 Print N (default max 10) time points from start point
     isodatetime R/2020/P1Y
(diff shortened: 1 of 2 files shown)


metomi/isodatetime/main.py
```diff
@@ -81,6 +76,10 @@
        The --max=N (default=10) can be used to control the maximum number
        of time points to print in the result.
 
+    Note: Negative durations can be provided in the following ways:
+        \\-PT1H
+        \\\\-PT1H
```
oliver-sanders: Note to reviewers:

```python
>>> print('\\')
\
```
```

File: metomi/isodatetime/main.py [hunk 2]
```diff
@@ -49,12 +49,7 @@
 
     # 3.  Print ISO8601 duration as total amount of a unit
```
wxtim: Suggestion
> #3. Print ISO8601 duration as a number of of hours (H), minutes (M) or second (S)

do we need the 3.1 line given that there is only one section here?

Is there a reason that you have got rid fo the examples?

metomi/isodatetime/main.py
```diff
@@ -49,12 +49,7 @@
...
     # 3.1 Into the total number of hours (H), minutes (M) or seconds (S)
-    #     it represents. Note: negative durations should be escaped by telling
-    #     the command to stop processing more options with `--` or by adding a
-    #     backslash in front of the duration.
-    isodatetime --as-total=s PT1H
-    isodatetime --as-total=s -- -PT1H
-    isodatetime --as-total=s \\-PT1H
+    #     it represents.
```
wxtim: I don't think that there is a test for this functionality - but that's out of scope for this issue.

wxtim:
Happy. 👍

metomi/isodatetime/main.py
datamel: Thanks Tim, Ive accepted those changes and put back the example that wasn't a negative duration - the negative durations have been added a bit further down. 
I added a test for checking the negative durations and these are all passing :)

PR #157, opened 2020-05-13
Ensure TimePoints are not init'd with out of bounds values

### LINKED ISSUE #18: Bounds checking for TimePoint class
The TimePoint class needs to do bounds checking for e.g. `day_of_month`, etc. This needs to be done here rather than in the parser, partly because of allowing different calendar formats (#16), and partly because it can be non-trivial.

This could perhaps be done by using the `self._tick_over()` method and seeing if any properties change - if they do, they were too high or too low.

### LINKED ISSUE #65: Normal time point + out of bounds truncated time point causes a loop forever

### LINKED ISSUE #155: Use of TimePoint in parsing datetime-like duration format
When using the datetime-like duration format, e.g. `P0004-03-10` for `P4Y3M10D`, but without specifying the month or the day of month, it sets the month and day to 1
```python
from metomi.isodatetime.parsers import DurationParser
parser = DurationParser()
print(parser.parse('P0004'))
# P4Y1M1D
```
This is because `DurationParser.parse()` parses the duration (without the P at the start) as a TimePoint, and TimePoint automatically sets the date to 1st Jan if the month & day are not specified

MetRonnie:
Do not allow stuff like `2020-00-00` or `2020-13-32T25:61` or `--02-30`
- Closes #18 
- Closes #65 

but allow unbounded values if using TimePoint for datetime-like durations
- Closes #155 

---
#### Also:

Check for conflicting inputs e.g. illegally specifying both `day_of_year` and `month_of_day|day_of_month`.

Fix 1 or 2 unnoticed, minor TimePoint bugs

---
#### Plus:

Reorganise tests somewhat, because `test_00.py` is bloated. Move some data model tests to new `test_01.py`. However, a lot of the untouched tests are intertwined so moving them would result in cylic dependencies.

When reviewing I suggest first looking at [the changes I made before](https://github.com/metomi/isodatetime/pull/157/files/87f8a0aa5c23fd9f0889ef6aaaeb18e0b75d005c) reorganising the tests. Then look at the [reorganisation of tests changes](https://github.com/metomi/isodatetime/pull/157/files/87f8a0aa5c23fd9f0889ef6aaaeb18e0b75d005c..90a324a9b8ad0403ea79c97d8b7939b4f406c2d0).


metomi/isodatetime/parsers.py
```diff
@@ -243,25 +241,12 @@ def parse(self, timepoint_string, dump_format=None, dump_as_parsed=False,
...
-            if changed_fields:
-                raise ValueError(
-                    'Invalid date-time components: ' +
-                    ', '.join((
-                        '%s=%s' % (key, before)
-                        for key, (before, after) in changed_fields.items()
-                        if before > after
-                    ))
-                )
-
```
MetRonnie, line 260: Remove validation in parser as superseded by bounds checking in TimePoint init

metomi/isodatetime/data.py
```diff
@@ -859,15 +885,34 @@ def __init__(self, expanded_year_digits=0, year=None, month_of_year=None,
...
-            if (self.month_of_year is None and self.week_of_year is None and
-                    self.day_of_year is None):
-                self.month_of_year = 1
-            if self.month_of_year is not None and self.day_of_month is None:
-                self.day_of_month = 1
-            if self.week_of_year is not None and self.day_of_week is None:
-                self.day_of_week = 1
+
+        specified_month = self.month_of_year or self.day_of_month
+        specified_week = self.week_of_year or self.day_of_week
```
kinow, line 890: Thought it was strange the specified_month to be the day_of_month, but looks like these variables are for validating the boundaries only, so no issue here I believe.

MetRonnie: I could change them to `is_month_specified = any((self.month_of_year, self.day_of_month))` etc if that makes more sense

kinow: I think that's fine :+1: these variables are used only in the code right after their declaration. So not too hard to see what they are intended for.

PR #161, opened 2020-06-01
Fix Codecov source path

### LINKED ISSUE #160: fix coverage
Coverage seems to be broken, last report received one year ago - https://codecov.io/gh/metomi/isodatetime/commits.

MetRonnie:
Closes #160

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -43,7 +43,7 @@ omit=
 parallel = False
 plugins=
 source=
-    ./isodatetime
+    ./metomi/isodatetime
 timid = False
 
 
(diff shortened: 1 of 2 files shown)


MetRonnie:
(I cancelled all but the `RUN_COVERAGE=true` jobs on Travis so it would finish quicker)

MetRonnie:
Turns out they didn't cancel and eventually did run anyway (apart from Mac). But it passed really: https://travis-ci.org/github/metomi/isodatetime/builds/693383068

oliver-sanders:
Codecov looks ok, happy to put this through with one approval since it's so small.

PR #163, opened 2020-06-05
Speed up non-coverage test jobs

MetRonnie:
Only use `coverage run setup.py` for the one `$RUN_COVERAGE == true` job, else just use `python setup.py`.

Approx times in minutes:

| Strategy          | Before | #161 | Now |
| ----------------- | ------ | ---- | --- |
| Ubuntu            | 10     | 25   | 5   |
| Ubuntu + coverage | 10     | 25   | 25  |
| Mac               | 30     | 50   | 20  |

(my previous PR #161 fixed the path of the isodatetime module in `.coveragerc`)

Travis build: https://travis-ci.org/github/metomi/isodatetime/builds/695070797

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,14 +51,19 @@ before_script:
 
 script:
   - |
-    export COVERAGE_PROCESS_START="${TRAVIS_BUILD_DIR}/.coveragerc"
-    export COVERAGE_FILE="${TRAVIS_BUILD_DIR}/.coverage"
+    if "$RUN_COVERAGE"; then
+      PY_CMD="coverage run"
+      export COVERAGE_PROCESS_START="${TRAVIS_BUILD_DIR}/.coveragerc"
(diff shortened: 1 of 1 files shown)


oliver-sanders:
WOW, that's how much coverage run slows things down!

oliver-sanders:
Tests don't appear to have run?!

MetRonnie:
Not sure why Travis isn't showing up in the checks, but it definitely ran, see the link I added above

oliver-sanders:
Travis is showing a different commit number for some reason.

MetRonnie:
Strange... it looks like it's running what you would get if you merged into master, rather than running the tip of the PR branch

oliver-sanders:
I can't see how this could break things so lets just merge it an see what happens.

MetRonnie:
Here's the build for the merge: https://travis-ci.org/github/metomi/isodatetime/builds/695110955

PR #164, opened 2020-06-05
travis: run tests in 4 processes

oliver-sanders:
Might a well give this a go whilst we are here.

This change allows you to run the really slow test using as many processes as you want.


MetRonnie:
Looks like it's only useful for making the coverage job run as quick as the non-coverage ones

MetRonnie:
I've put up a PR for only running in 4 processes when it's the coverage test https://github.com/oliver-sanders/isodatetime/pull/1

oliver-sanders:
> I've put up a PR on this PR branch for only running in 4 processes when it's the coverage test

Not sure why we would want to do that?

oliver-sanders:
There wasn't much of a speedup because the slow test was monolithic so:

* I've split it up using pytest so that xdist can distribute the problem
* I've got rid of the threadpoolexecutor because we need processes not threads
* I got rid of the `--runslow` infrastructure in the process, use `-m slow` or `-m 'not slow'`.

oliver-sanders:
(will see what the performance looks like before disabling slow tests on selected platforms)

MetRonnie:
> Not sure why we would want to do that?

It did at least speed up the coverage run from 25 mins to 5 mins, but didn't seem to affect the others (also Python 3.7 and below seems to crash with it enabled). Not sure it's worth trying to improve 5 mins as that's pretty good I'd have thought

oliver-sanders:
The Python<3.7 crash is due to Travis not installing the testing dependencies correctly, should be an easy fix.

Looks like the entire battery should now pass in ~5mins

.travis.yml
```diff
@@ -38,46 +38,23 @@ install:
...
-    # NOTE: the way to handle long-living commands in the documentation is
-    # travis_wait 40 coverage run -m isodatetime.tests.test_time_point
-    # but it does not produce any output, which can be frustrating in case of errors.
-    # So here we are using the approach documented in the issue
-    # https://github.com/travis-ci/travis-ci/issues/4190.
-    $PY_CMD setup.py test --pytest-args='--runslow' &
-    TEST_PID="$!"
-    minutes=0
-    limit=50
-    while ps -p "${TEST_PID}" >/dev/null 2>&1; do
```
oliver-sanders, line 70: This stuff was only needed because of the long-running monolithic test, now that its been broken apart into chunks this is no longer necessary.

conftest.py
```diff
@@ -1,39 +0,0 @@
...
-"""Local conftest.py plugins contain directory-specific hook implementations.
-Session and test running activities will invoke all hooks defined in
-conftest.py files closer to the root of the filesystem."""
-
-import pytest
-
-
-def pytest_addoption(parser):
-    parser.addoption(
-        "--runslow", action="store_true", default=False, help="run slow tests"
```
oliver-sanders, line 28: pytest has a built-in system for marking tests so we don't have to implement our own.

setup.py
```diff
@@ -70,20 +53,29 @@ def run(self):
...
 
-# Only include pytest-runner in setup_requires if we're invoking tests
-if {'pytest', 'test', 'ptr'}.intersection(sys.argv):
-    setup_requires = ['pytest-runner']
-else:
-    setup_requires = []
+install_requires = []
+tests_require = [
+    'pytest>=5',
+    'pytest-env'
```
oliver-sanders, line 59: Removed `pytest-cov` as it isn't a testing dependency, it is only needed for codecov upload.

setup.py
```diff
@@ -70,20 +53,29 @@ def run(self):
         super().run()
 
 
-# Only include pytest-runner in setup_requires if we're invoking tests
-if {'pytest', 'test', 'ptr'}.intersection(sys.argv):
-    setup_requires = ['pytest-runner']
```
oliver-sanders, line 75: Removed the `pytest-runner` as you can just run `pytest` on the command line.

MetRonnie:
Why is the coverage test back up to 25 minutes? My only other concern is that the log on Travis is now very long and it looks like it might be difficult to figure out which test(s) have failed.

oliver-sanders:
> Why is the coverage test back up to 25 minutes?

Dam, has that gone up? What was it before?

> My only other concern is that the log on Travis is now very long and it looks like it might be difficult to figure out which test(s) have failed.

The log is the same length as before no?

As with Cylc Flow, scroll to the bottom to see the results from pytest (BTW I don't like the pytest reporting format but that's a different matter).

MetRonnie:
| Strategy          | Before this PR | First run of this PR | Latest run of this PR |
| ----------------- | -------------- | -------------------- | --------------------- |
| Ubuntu            | 5:30           | 4:30                 | 4:30                  |
| Ubuntu + coverage | 26:30          | 4:30                 | 25:00                 |
| Mac               | 19:30          | 15:00                | 15:30                 |

MetRonnie:
According to the [docs](https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources), each GH Actions VM only has 2 CPUs, so I'd have thought this means 2 processes are more appropriate than 4

.coveragerc
```diff
@@ -16,65 +16,21 @@
...
-#    trace
-# Include can be used only if source is not used!
-#include
-note=
-omit=
-    */isodatetime/test*
-parallel = False
-plugins=
+omit=s
+    metomi/isodatetime/tests
```
MetRonnie, line 22: `omit=s` possible typo?

Issue #50, opened 2014-07-08
Duration comparison: be careful about badly-defined unit comparison


benfitzpatrick:
We should improve the way that comparisons are done between durations with badly defined units like months and years.

PR #165, opened 2020-06-08
Make data classes hashable

### LINKED ISSUE #162: make data classes hashable
Perhaps the biggest architectural limitation on isodatetime is the fact that the data classes (e.g. TimePoint, Duration, TimeRecurrence) are mutable objects.

They should all be immutable and hashable, this would widen the scope for caching and efficiency improvements.

The biggest barrier to achieving this are some of the methods on `TimePoint` which mutate the `TimePoint` itself (e.g. `to_week_date`), the easy solution is to return a new object rather than mutating the existing one.

Once done we can `lru_cache` `TimeRecurrence.get_is_valid` which is currently the biggest offender in Cylc.

MetRonnie:
Close #162 (plus #166 and #50).

As noted in the discussion below, for the hash of two instances to be the same, the `==` operator has to evaluate as True. This is tricky for `Duration`s with nominal units, as `P1Y != P365D` (explained in #167).

The behaviour settled on, which is not ideal, is to have `P1Y` tacitly equal to 365 days (or rather `CALENDAR.ROUGH_DAYS_IN_YEAR`) for the sake of the `<`, `<=`, `>` etc operators (this behaviour is pretty much unchanged from before), but not actually explicitly equal when it comes to `==`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,11 +302,11 @@ years, months, days, hours, minutes, and seconds are used literally
 #### 1 - Recur with a duration given by the difference between a start date
 and a subsequent date, starting at the start date
 
-Example Syntax           | Example                 | Meaning
- ----------------------- | ----------------------- | ------------------------------------------------------------------
-R/CCYY/CCYY              | R/2010/2014             | Repeat every 4 years, starting at 2010-01-01.
-R/CCYY-MM/CCYY-DDD       | R/2010-01/2012-045      | Repeat every 2 years and 44 days, starting at 2010-01-01
-R5/CCYY-Www-D/CCYY-Www-D | R/2015-W05-2/2015-W07-3 | Repeat every 2 weeks and 1 day, five times, starting at 2015-W05-2
+Example Syntax           | Example                  | Meaning
+ ----------------------- | ------------------------ | ------------------------------------------------------------------
+R/CCYY/CCYY              | R/2010/2014              | Repeat every 4 years, starting at 2010-01-01.
+R/CCYY-MM/CCYY-DDD       | R/2010-01/2012-045       | Repeat every 2 years and 44 days, starting at 2010-01-01
+Rn/CCYY-Www-D/CCYY-Www-D | R5/2015-W05-2/2015-W07-3 | Repeat every 2 weeks and 1 day, five times, starting at 2015-W05-2
 
 #### 2 - Recur with a given duration, starting at a context date-time
 
--- a/README.md
+++ b/README.md
@@ -47,12 +47,12 @@ Python API:
 >>> duration.get_days_and_seconds()
 (365.0, 10800.0)
 >>> date_time + duration
-2001-01-01T03:00:00Z
+<metomi.isodatetime.data.TimePoint: 2001-01-01T03:00:00Z>
 
 # Recurrences
 >>> recurrence = parse.TimeRecurrenceParser().parse('R/1999/P1Y')
 >>> recurrence.get_next(date_time)
-2001-01-01T00:00:00Z
+<metomi.isodatetime.data.TimePoint: 2001-01-01T00:00:00Z>
 
 # Output
 >>> dump.TimePointDumper().strftime(date_time, '%d/%M/%Y %H:%M:%S')
--- a/README.md
+++ b/README.md
@@ -91,9 +91,9 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 You should have received a copy of the GNU Lesser General Public License along
 with this program.  If not, see [GNU licenses](http://www.gnu.org/licenses/).
 
-## ISO8601
+## ISO 8601
 
-[ISO8601 (2004)](https://www.iso.org/standard/40874.html)
+[ISO 8601 (2004)](https://www.iso.org/standard/40874.html)
 is an international standard for writing down date/time information.
 
 It is the correct, internationally-friendly, computer-sortable way to
@@ -277,6 +277,20 @@ Syntax    | Example  | Meaning
 
 Combining any other unit with weeks is not allowed. Decimals may only be used
 for hours, minutes and seconds.
+
+Note that years and months are "nominal" durations, whose exact length of time
+depends on their position in the calendar. E.g., a duration of 1 calendar year
+starts on a particular day of a particular month and ends on the same day of
+the same month in the following calendar year, and may be different to 365 days
+in the Gregorian calendar due to leap years.
+
+Conversely, weeks, days, hours, minutes and seconds are exact units, so
+`P1W == P7D`, `P1D == PT24H` and `PT1H == PT60M` etc. are always true.
+(Although ISO 8601 specifies that weeks and days are nominal durations, there
+is no case where they are not exact in our implementation.)
+<!-- ...because TimePoints always have time zones assigned to them (apart
+from truncated TimePoints, but you can't add Durations to truncated
+TimePoints). Local time zones don't actually exist in our implementation. -->
 
 A supplementary format (which has to be agreed in advance) is to specify a
 date-time-like duration (`PCCYY-MM-DDThh:mm:ss`) where the numbers given for
--- a/README.md
+++ b/README.md
@@ -299,8 +299,7 @@ years, months, days, hours, minutes, and seconds are used literally
 
 ### Recurring date-time series
 
-#### 1 - Recur with a duration given by the difference between a start date
-and a subsequent date, starting at the start date
+#### 1. Recur with a duration given by the difference between a start date and a subsequent date
 
 Example Syntax           | Example                  | Meaning
  ----------------------- | ------------------------ | ------------------------------------------------------------------
@@ -308,7 +307,7 @@ R/CCYY/CCYY              | R/2010/2014              | Repeat every 4 years, star
 R/CCYY-MM/CCYY-DDD       | R/2010-01/2012-045       | Repeat every 2 years and 44 days, starting at 2010-01-01
 Rn/CCYY-Www-D/CCYY-Www-D | R5/2015-W05-2/2015-W07-3 | Repeat every 2 weeks and 1 day, five times, starting at 2015-W05-2
 
-#### 2 - Recur with a given duration, starting at a context date-time
+#### 2. Recur with a specified duration, starting at a context date-time
 
 (You have to supply the context somewhere else)
 
@@ -317,7 +316,7 @@ Example Syntax        | Example          | Meaning
 R/PnMnDTnM            | R/P10M3DT45M     | Repeat every 10 months, 3 days, and 45 minutes from a context start date-time.
 Rn/PnY                | R2/P4Y           | Repeat every 4 years, for a total of 2 times, from a context start date-time.
 
-#### 3 - Recur with a given duration starting at a particular date-time
+#### 3. Recur with a specified duration starting at a particular date-time
 
 Example Syntax             | Example                  | Meaning
  ------------------------- | ------------------------ | ----------------------------------------------------------------------------------------------
@@ -326,14 +325,17 @@ R/CCYY-Www-D/PnW           | R/2012-W02-1/P1W         | Repeat weekly starting a
 R/CCYYDDDThhmm/PnD         | R/1996291T0630+0100/P2D  | Repeat every 2 days starting on the 291st day of 1996 at 06:30, UTC + 1
 Rn/CCYY-MM-DDThh:mm/PTnH   | R2/19900201T06Z/PT12H    | Repeat every 12 hours, for a total of 2 repetitions, starting at 1990-02-01T06Z
 Rn/CCYY-Www-D/PnW          | R5/2012-W02-1/P1W        | Repeat weekly, for a total of 5 repetitions, starting at Monday in the second ISO week of 2012
-Rn/CCYYDDDThhmm/PnD        | R1/1996291T0630+0100/P2D | Repeat once at the 291st day of 1996 at 06:30, UTC + 1
+Rn/CCYYDDDThhmm/PnD        | R1/1996291T0630+0100/P2D | Repeat once at the 291st day of 1996 at 06:30, UTC + 1 (note the duration is ignored)
 
-#### 4 - Recur with a given duration counting back from a particular date-time
+#### 4. Recur with a specified duration ending at a particular date-time
+
+The starting date-time of the recurrence is calculated from the specified
+duration.
 
 Example Syntax             | Example                  | Meaning
  ------------------------- | ------------------------ | ---------------------------------------------------------------
-R/PTnH/CCYY-MM-DDThhZ      | R/PT1H/2012-01-02T00Z    | Repeat hourly counting back from 2012-01-02T00Z
-R/PnY/CCYY                 | R/P3Y/2000               | Repeat every 3 years counting back from 2000-01-01.
-R/PTnS/+XCCYYDDDThhmm      | R/PT5s/-002500012T1800   | Repeat every 5 seconds counting back from the 12th day in 2501 BC at 18:00 (using 2 expanded year digits).
-Rn/PnYTnM/CCYY-MM-DDThhZ   | R5/P1YT5M/2012-01-02T00Z | Repeat every year and 5 minutes counting back from 2012-01-02T00Z
-Rn/PnM/CCYY-MM             | R4/P1M/2000-05           | Repeat monthly, four times, counting back from 2000-05-01.
+R/PTnH/CCYY-MM-DDThhZ      | R/PT1H/2012-01-02T00Z    | Repeat hourly until 2012-01-02T00Z
+R/PnY/CCYY                 | R/P3Y/2000               | Repeat every 3 years until 2000-01-01.
+R/PTnS/+XCCYYDDDThhmm      | R/PT5s/-002500012T1800   | Repeat every 5 seconds until the 12th day in 2501 BC at 18:00 (using 2 expanded year digits).
+Rn/PnYTnM/CCYY-MM-DDThhZ   | R5/P1YT5M/2012-01-02T00Z | Repeat every year and 5 minutes, five times, until 2012-01-02T00Z
+Rn/PnM/CCYY-MM             | R4/P1M/2000-05           | Repeat monthly, four times, until 2000-05-01.
--- a/metomi/isodatetime/data.py
+++ b/metomi/isodatetime/data.py
@@ -377,9 +377,8 @@ class Duration(object):
         self._hours = hours
         self._minutes = minutes
         self._seconds = seconds
-        if (not self.years and not self.months and not self.hours and
-                not self.minutes and not self.seconds and
-                weeks and not days):
+        if (weeks and not years and not months and not days and
+                not hours and not minutes and not seconds):
             self._weeks = self.days // CALENDAR.DAYS_IN_WEEK
             self._years, self._months, self._days = (None, None, None)
             self._hours, self._minutes, self._seconds = (None, None, None)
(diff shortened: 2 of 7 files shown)


metomi/isodatetime/data.py
```diff
@@ -522,54 +542,50 @@ def __floordiv__(self, other):
...
 
+    def __hash__(self) -> int:
+        return hash(self.get_days_and_seconds())
+
     def __eq__(self, other: "Duration") -> bool:
-        my_data = self.get_days_and_seconds()
-        other_data = other.get_days_and_seconds()
-        return my_data == other_data
+        # TODO: check instance of other
+        return self.get_days_and_seconds() == other.get_days_and_seconds()
```
MetRonnie, line 563: A possible problem with this is that months & years are ambiguous lengths of time. Nevertheless, the existing implementation already meant that e.g. `P1M == P30D`.

metomi/isodatetime/data.py
```diff
@@ -522,54 +542,50 @@ def __floordiv__(self, other):
...
+        new._years //= other
+        new._months //= other
+        new._days //= other
+        new._hours //= other
+        new._minutes //= other
+        new._seconds //= other
         return new
 
+    def __hash__(self) -> int:
+        return hash(self.get_days_and_seconds())
```
oliver-sanders, line 558: Because `get_days_and_seconds` duplicates the instance it would probably be more efficient to extract the required data into a tuple.

```python
return hash((getattr(self, attr) for attr in self.DATA_ATTRIBUTES))
```

metomi/isodatetime/data.py
oliver-sanders: No problem, it's what the code did before and I think Python's `__eq__` is meant to be more like `=` than `≡` anyway.

metomi/isodatetime/data.py
```diff
@@ -522,54 +542,50 @@ def __floordiv__(self, other):
...
-        return my_data == other_data
+        # TODO: check instance of other
+        return self.get_days_and_seconds() == other.get_days_and_seconds()
 
     def __ne__(self, other: "Duration") -> bool:
         return not self.__eq__(other)
 
     def __lt__(self, other: "Duration") -> bool:
-        my_data = self.get_days_and_seconds()
-        other_data = other.get_days_and_seconds()
```
oliver-sanders, line 547: Cutting out these unnecessary assignments is a worthwhile improvement, thanks.

metomi/isodatetime/data.py
```diff
@@ -522,54 +542,50 @@ def __floordiv__(self, other):
...
-        for attr in ["years", "months", "weeks", "days", "hours",
-                     "minutes", "seconds"]:
+        for attr in self.DATA_ATTRIBUTES:
             if getattr(self, attr, None):
                 return True
         return False
 
     def __str__(self):
+        if not self:
+            return "P0Y"
```
oliver-sanders, line 587: Is this a bug fix?

metomi/isodatetime/data.py
```diff
@@ -413,11 +433,9 @@ def get_days_and_seconds(self):
         Duration which has self.seconds = CALENDAR.SECONDS_IN_DAY +
         100 will return 1 day, 100 seconds or (1, 100) from this
         method.
-
```
oliver-sanders, line 416: Doesn't matter but FYI these trailing blank-lines in docstrings are legal - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

metomi/isodatetime/data.py
MetRonnie: Just refactoring, previously it would go through a whole set of logic and only then to effectively check if any of the properties were not None

metomi/isodatetime/data.py
MetRonnie: The problem then is it would mean `PT1H != PT60M`

oliver-sanders: I think for a `__hash__` that is the correct answer.

MetRonnie: However I have just spotted ways that `get_days_and_seconds()` can be optimised

MetRonnie: 46eb487 makes it 10x faster when the Duration is in weeks, but can't really speed up otherwise

MetRonnie:
(The tests fail because of things like TimeZone inheriting from Duration. I'm checking the Duration tests pass locally as I go along)

metomi/isodatetime/data.py
MetRonnie: > I think for a __hash__ that is the correct answer.

According to [the docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__):
> The only required property is that objects which compare equal have the same hash value

oliver-sanders: Having two instances with different attributes producing the same hash kinda breaks the contract of `__hash__`.

We can't have `P1Y` and `P365D` producing the same hash as this will cause issues with caching logic e.g:

```python
# untested

from functools import lru_cache

@lru_cahe
def add(this, that):
    return this + that

t1 = TimePoint('2000')
d1 = Duration('P1Y')
d2 = Duration('P365D') 

# do this one way
add(t1, d1)  # 2001-01-01
add(t1, d2)  # 2001-01-01

# but do it the other
add.clear()
add(t1, d2)  # 2000-12-31
add(t1, d1)  # 2000-12-31
```

The nicest way out of this might be to change the behaviour of `__eq__`, I'm trying to work out what edge cases this actually applies to:

```python
def __eq__(self, other):
    if (
            # cases for which get_days_and_seconds is not accurate
            self.years != other.years
            and self.months != other.months
            # ...
    ):
        return False
    return self.get_days_and_seconds() == other.get_days_and_seconds()
```

MetRonnie: I don't have an answer yet, but having revisited the rationale behind doing this, I think the main criterion for the hash being the same has to be whether the methods will return the same results.

And I agree that `x` years or months can't be considered the same as `y` days

(except for the case of the 360-day calendar, and years in the 365/6-day calendars, ugh)

oliver-sanders: > ugh

🤢

MetRonnie: Another thing: `P1D` is not necessarily `PT24H`! Because there might be a daylight saving change overnight

oliver-sanders: Good point, so why did we try to determine equality.

MetRonnie: At least in the case of hours, minutes & seconds, they are always exact and equality definitely applies for them (having reviewed the 8601 standard - [draft spec here](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf))

MetRonnie:
Given that we think `P365D == P1Y` should be `False`, what about `P365D > P1Y` or `P365D < P1Y`?

`raise ValueError()`?

oliver-sanders:
> my brain hurts

Why are dates so a hard!

I think it's ok if these comparison methods come with the caveat that they operate under certain assumptions. Accurate comparisons require sensical entities to compare. I would be happy to extend that to `__eq__`, it's just that `__hash__` gets in the way!

How about just changing the rule for equality and leaving the rest as it is (with a warning) since there is no correct solution:

```python
P365D != P1Y
P1Y == P1Y
P365D < P1Y
P1Y <= P1Y
```

wxtim:
> Why are dates so a hard!
Did you remove the stone?

MetRonnie:
> How about just changing the rule for equality and leaving the rest as it is (with a warning) since there is no correct solution

When you say warning is that `warnings.warn()` or just leave a note of caution in the docstring?

oliver-sanders:
Caution on the docstring I think. A formal warning is a bit over the top.

MetRonnie:
Problem with making TimePoint hashable: TimePoints representing the same point in time but differing in time zone and whether they're in calendar date mode, ordinal date mode or week date mode all evaluate as equal

oliver-sanders:
> But the methods can return different values for each of these.

*[insert cathartic scream]*

Oh god, and it seemed so simple on the surface.

If it's that bad we might want to consider another approach, perhaps we should implement new equality methods which use a more loosely defined equality e.g. `TimePoint.equivalent_to` or something like that. But make `__eq__` strict as per the pythonic definition.

The good news is that Cylc hides the isodatetime classes behind its cycling classes. Whilst this is a bad architectural decision it does at least mean that the code base isn't doing comparisons on raw isodatetime objects so the impact of this to Cylc is likely to be minimal.

MetRonnie:
For methods that depend on "variable" properties that don't affect the hash of the instance, could we just create a wrapper method that takes the variable properties as args?

Thus, two instances of different time zones, but representing the same point when converted to UTC, would be equal but the method to get the hours, minutes and seconds wouldn't get cached erroneously?

Something like this:
```python
class TimePoint:
    ...
    def get_hours_mins_secs(self):
        return self._get_hours_mins_secs(self.time_zone)

    @lru_cache(1000)
    def _get_hours_mins_secs(self, _time_zone):
        return (self.hour_of_day, self.minute_of_hour, self.second_of_minute)
```

MetRonnie:
Without caching, the performance has gone down

`test_time_point.py`, before:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2198    0.185    0.000    4.826    0.002 test_time_point.py:72(_do_test_dates)
    19782    0.053    0.000    2.154    0.000 data.py:1522(__sub__)
   153818    0.268    0.000    1.879    0.000 data.py:357(__init__)
   156016    1.105    0.000    1.630    0.000 data.py:2416(_type_checker)
    35168    0.101    0.000    1.285    0.000 data.py:1280(__add__)
    87920    0.079    0.000    1.216    0.000 data.py:646(copy)
    90118    0.075    0.000    1.168    0.000 data.py:642(__init__)
    43960    0.218    0.000    0.968    0.000 data.py:1361(copy)
     4396    0.024    0.000    0.755    0.000 data.py:1466(__gt__)
    37366    0.027    0.000    0.622    0.000 case.py:847(assertEqual)
    46158    0.510    0.000    0.562    0.000 data.py:2070(get_calendar_date_from_week_date)
    17584    0.154    0.000    0.505    0.000 data.py:1369(get_props)
    19782    0.021    0.000    0.477    0.000 data.py:966(get_ordinal_date)
     8792    0.005    0.000    0.467    0.000 case.py:840(_baseAssertEqual)
    19782    0.013    0.000    0.447    0.000 data.py:2135(get_ordinal_date_from_week_date)
    24178    0.078    0.000    0.424    0.000 data.py:508(__mul__)
     4396    0.011    0.000    0.405    0.000 data.py:1379(__eq__)
    26376    0.031    0.000    0.375    0.000 data.py:939(get_calendar_date)
    56930    0.271    0.000    0.347    0.000 data.py:1601(tick_over)
     8792    0.012    0.000    0.319    0.000 data.py:1111(set_time_zone)
     8792    0.009    0.000    0.303    0.000 data.py:505(__sub__)

```

After:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     2198    0.191    0.000    9.112    0.004 test_time_point.py:72(_do_test_dates)
    43960    0.175    0.000    6.853    0.000 data.py:1467(__add__)
    56940    0.441    0.000    6.042    0.000 data.py:1716(_tick_over)
    56940    3.763    0.000    5.452    0.000 data.py:1800(_tick_over_day_of_month)
    19782    0.047    0.000    3.599    0.000 data.py:1635(__sub__)
 21535779    1.618    0.000    1.618    0.000 data.py:1085(day_of_month)
    59294    0.089    0.000    0.616    0.000 data.py:421(__init__)
     8792    0.042    0.000    0.616    0.000 data.py:1568(_cmp)
    41762    0.367    0.000    0.545    0.000 data.py:2519(_type_checker)
    46158    0.222    0.000    0.544    0.000 data.py:1540(_copy)
   101056    0.256    0.000    0.416    0.000 data.py:493(_copy)
    37366    0.027    0.000    0.406    0.000 case.py:847(assertEqual)
     4396    0.004    0.000    0.395    0.000 data.py:1629(__gt__)
    17584    0.175    0.000    0.346    0.000 data.py:1548(get_props)
     8792    0.014    0.000    0.308    0.000 data.py:1290(to_time_zone)
     8792    0.005    0.000    0.249    0.000 case.py:840(_baseAssertEqual)
     4396    0.002    0.000    0.227    0.000 data.py:1620(__eq__)
  2236984    0.214    0.000    0.214    0.000 {built-in method builtins.getattr}
    24178    0.070    0.000    0.205    0.000 data.py:616(__mul__)
     8792    0.008    0.000    0.160    0.000 data.py:613(__sub__)
  1654601    0.156    0.000    0.156    0.000 {built-in method builtins.setattr}
```

metomi/isodatetime/data.py
```diff
@@ -1060,111 +1254,108 @@ def get(self, property_name):
...
-        if offset.minutes:
-            if self.minute_of_hour is None:
-                self.hour_of_day += (
-                    offset.minutes / float(CALENDAR.MINUTES_IN_HOUR))
-            else:
-                self.minute_of_hour += offset.minutes
-            self.tick_over()
-        if offset.hours:
-            self.hour_of_day += offset.hours
-            self.tick_over()
```
MetRonnie, line 1103: Not needed, can just add TimeZone to the TimePoint instead

MetRonnie:
Forgot to put [skip travis] in update changelog commit, cancelled the build

metomi/isodatetime/data.py
```diff
@@ -302,33 +343,57 @@ def __iter__(self):
...
+        if self._format_number == 1:
+            return prefix + str(self._start_point) + "/" + str(self._end_point)
+        elif self._format_number == 3:
+            return prefix + str(self._start_point) + "/" + str(self._duration)
+        elif self._format_number == 4:
+            return prefix + str(self._duration) + "/" + str(self._end_point)
         return "R/?/?"
 
+    def __repr__(self):
+        return "<isodatetime.data.TimeRecurrence: {0}>".format(repr(str(self)))
```
kinow, line 375: Maybe we should use something like (not tested) `f"<{self.__class__.__module__}.{self.__class__.__name__}: {repr(str(self))}}>`, just in case it gets moved/renamed?

metomi/isodatetime/data.py
```diff
@@ -365,50 +433,77 @@ def __init__(self, years=0, months=0, weeks=0, days=0,
...
-                weeks and not days):
-            self.weeks = self.days // CALENDAR.DAYS_IN_WEEK
-            self.years, self.months, self.days = (None, None, None)
-            self.hours, self.minutes, self.seconds = (None, None, None)
+                self._days += CALENDAR.DAYS_IN_WEEK * weeks
+        self._hours = hours
+        self._minutes = minutes
+        self._seconds = seconds
+        if (weeks and not years and not months and not days and
+                not hours and not minutes and not seconds):
```
kinow, line 449: Much easier to read with the positive value first, followed by all negations :+1:

metomi/isodatetime/data.py
```diff
@@ -442,57 +536,73 @@ def get_seconds(self):
...
+            new._days = new._weeks * CALENDAR.DAYS_IN_WEEK
+            new._weeks = None
+            return new
+        return self
 
     def to_weeks(self):
-        """Convert to week representation (warning: use with caution)."""
+        """Return a new Duration in week representation (use with caution -
+        this returns the floor of the decimal number of weeks, so might lose
+        precision)."""
```
kinow, line 576: :clap: kudos for explaining why!

metomi/isodatetime/data.py
```diff
@@ -612,58 +738,90 @@ def __str__(self):
...
-        elif content_string.endswith("T"):
+        if content_string.endswith("T"):
             # No time unit information, so strip the delimiter.
             content_string = content_string[:-1]
 
         total_string = start_string + content_string
         return total_string.replace(".", ",")
 
+    def __repr__(self):
+        return "<isodatetime.data.Duration: {0}>".format(repr(str(self)))
```
kinow, line 749: Use `self` values instead of hard-coded values maybe? (see similar comment above)

metomi/isodatetime/data.py
```diff
@@ -612,58 +738,90 @@ def __str__(self):
...
             time_string = "+%02d:%02d"
-            if self.hours < 0 or (self.hours == 0 and self.minutes < 0):
+            if self._hours < 0 or (self._hours == 0 and self._minutes < 0):
                 time_string = "-%02d:%02d"
-            return time_string % (abs(self.hours), abs(self.minutes))
+            return time_string % (abs(self._hours), abs(self._minutes))
 
     def __repr__(self):
-        return "<isodatetime.data.TimeZone:" + repr(str(self)) + ">"
+        return "<isodatetime.data.TimeZone: {0}>".format(repr(str(self)))
```
kinow, line 821: See previous comments about using values from `__self__` instead of hard-coding? (Although thinking again now, this PR is about making it hashable, so I think we can move it for later)

metomi/isodatetime/data.py
```diff
@@ -1217,314 +1407,240 @@ def add_truncated(self, year_of_century=None, year_of_decade=None,
...
+                    if op == "eq":
+                        return False
+                    if op == "lt":
+                        return self_attr < other_attr
+                    if op == "le":
+                        return self_attr <= other_attr
+                    if op == "gt":
+                        return self_attr > other_attr
+                    if op == "ge":
+                        return self_attr >= other_attr
```
kinow, line 1602: I **think** later it could be possible to use the `operator` module? Something like have a dict of operators (e.g. `operator.lt.__name__ => operator.lt`and etc.  I think?). Then just `op = dict.get(op); return op(self_attr, other_attr)`.

README.md
```diff
@@ -47,12 +47,12 @@ Python API:
 >>> duration.get_days_and_seconds()
 (365.0, 10800.0)
 >>> date_time + duration
-2001-01-01T03:00:00Z
+<metomi.isodatetime.data.TimePoint: 2001-01-01T03:00:00Z>
```
MetRonnie, line 50: Is this new `__repr__` format too verbose?

metomi/isodatetime/data.py
oliver-sanders: Not needed but could iterate `self.__slots__` to make the `repr` reproducible e.g: `TimeRecurrence(years=1, months=1)`

README.md
oliver-sanders: I don't think it matters too much.

The original intension of `repr` was to be **repr**oducible but these days it's a bit of a free-for-all (I've sometimes miss-use them on-purpose), whatever works.

Other code shouldn't really be relying on the `repr` format so we should be free to change.

metomi/isodatetime/data.py
MetRonnie: Unfortunately f strings aren't supported by Python 3.5

metomi/isodatetime/data.py
```diff
@@ -531,60 +646,76 @@ def __floordiv__(self, other):
...
-    def __eq__(self, other: "Duration") -> bool:
-        my_data = self.get_days_and_seconds()
-        other_data = other.get_days_and_seconds()
-        return my_data == other_data
+    def __hash__(self) -> int:
+        # TODO: alt calendar modes
+        if self.get_is_in_weeks():
+            return hash((0, 0, self._get_non_nominal_seconds()))
+        return hash(
+            (self._years, self._months, self._get_non_nominal_seconds()))
```
oliver-sanders, line 666: Could this return the same value for a duration in weeks as one which isn't?

Since we are already performing the check, perhaps:

```python
return hash((self.get_is_in_weeks(), self._years, self._months, self._get_non_nominal_seconds()))
```

metomi/isodatetime/data.py
```diff
@@ -531,60 +646,76 @@ def __floordiv__(self, other):
...
+                    return (self._get_non_nominal_seconds() ==
+                            other._get_non_nominal_seconds())
+                return False
+            return (
+                self._years == other._years and
+                self._months == other._months and
+                self._get_non_nominal_seconds() ==
+                other._get_non_nominal_seconds()
+            )
+        return NotImplemented
```
oliver-sanders, line 681: Might be good to `return False` here otherwise we will have problems comparing to `None` instances.

Ok to raise exceptions in other cases though.

metomi/isodatetime/data.py
MetRonnie: `hash(P1W) == hash(P7D)`, seeing as `P1W == P7D`. Including `get_is_in_weeks` would break this behaviour.

metomi/isodatetime/data.py
MetRonnie: Returning `NotImplemented` delegates to `other`'s `__eq__` method, which for `None` seems to do something along the lines of
```python
return False if other is not None else True
```
So shouldn't be a problem

oliver-sanders: Sorry, I read `raise NotImplementedError` rather than `return NotImplemented`!

metomi/isodatetime/data.py
oliver-sanders: Er, um, having methods return different values for objects which hash the same isn't great, are there any other methods or properties which would return different values or is it just `get_is_in_weeks` (if so no biggie).

MetRonnie: I know it isn't great, presumably if we don't cache `get_is_in_weeks()` it should be okay though? `__str__()` is the only other one, unless you count e.g. `_copy()` and `__add__()` which return new `Duration`s, which may or may not be in weeks

oliver-sanders: Sorry for being forgetful (too many PRs going on at the moment), could you remind me why we want `hash(P1W) == hash(P7D)`.

MetRonnie: No problem, it's because [the docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__) say
> The only required property is that objects which compare equal have the same hash value

oliver-sanders: Thanks for reminding me!

Ok, lets just keep this in the back of our minds for now.

oliver-sanders:
(note I've renamed the milestone to 2.1 as this is a pretty significant change)

oliver-sanders:
Let's see how Cylc Flow handles this 😈

PR #167, opened 2020-06-09
Durations: do not allow decimals for years, months, weeks, days

### LINKED ISSUE #147: Duration ISO8601 - Allow decimal fraction for smallest of units
It is documented here that there could be support for decimal fraction in the latest value

> The smallest value used may also have a decimal fraction, as in "P0.5Y" to indicate half a year. This decimal fraction may be specified with either a comma or a full stop, as in "P0,5Y" or "P0.5Y". 

I have tested unsuccessfully:

```bash
File "/home/ljdm/.pyenv/versions/scheduled-table-deletion/lib/python3.7/site-packages/isodatetime/parsers.py", line 578, in parse
assumed_time_zone=(0, 0)
File "/home/ljdm/.pyenv/versions/scheduled-table-deletion/lib/python3.7/site-packages/isodatetime/parsers.py", line 603, in parse_timepoint_expression
return parser.parse(timepoint_expression)
File "/home/ljdm/.pyenv/versions/scheduled-table-deletion/lib/python3.7/site-packages/isodatetime/parsers.py", line 237, in parse
date_info, time_info, parsed_expr = self.get_info(timepoint_string)
File "/home/ljdm/.pyenv/versions/scheduled-table-deletion/lib/python3.7/site-packages/isodatetime/parsers.py", line 430, in get_info
keys, date_info = self.get_date_info(date)
File "/home/ljdm/.pyenv/versions/scheduled-table-deletion/lib/python3.7/site-packages/isodatetime/parsers.py", line 390, in get_date_info
raise ISO8601SyntaxError("date", date_string)
isodatetime.parsers.ISO8601SyntaxError: Invalid ISO 8601 date representation: 0.5Y
isodatetime.parsers.ISO8601SyntaxError: Invalid ISO 8601 duration representation: P0.5Y
isodatetime.parsers.ISO8601SyntaxError: Invalid ISO 8601 date representation: 0,5Y
isodatetime.parsers.ISO8601SyntaxError: Invalid ISO 8601 duration representation: P0,5Y
```

MetRonnie:
Close #147 

Years, months, weeks and days are nominal durations, which depend on their position in the calendar, unlike time durations (hours, mins, secs). A year is not necessarily equal to 365 or 366 days; instead it is the duration that starts at a particular time on a particular day of a particular month, and ends at the same time of the same day of the same month in the following calendar year.

Even 1 day is not necessarily equal to 24 hours if we are using a local time zone, due to the possibility of a daylight saving change. Instead 1 day is the duration that starts at a particular time of day and ends at the same time of day on the following calendar day.

It follows that decimals of these nominal durations are not easily defined. In order to define 0.5 years it seems you would have to calculate the number of hours between the start date and the same date a year later, then halve that number of hours.

We haven't actually implemented decimals for these nominal durations anyway, but up until now you could pass floats to them when initialising a `Duration`.

Sorry, that ended up a bit long. Rant over 😄

### DIFF
--- a/metomi/isodatetime/data.py
+++ b/metomi/isodatetime/data.py
@@ -2322,9 +2322,8 @@ def get_timepoint_from_seconds_since_unix_epoch(num_seconds, utc=False):
         **CALENDAR.UNIX_EPOCH_DATE_TIME_REFERENCE_PROPERTIES)
     reference_timepoint.set_time_zone_to_local()
     if utc:
-        adj_secs = reference_timepoint.time_zone.get_seconds()
+        num_seconds += reference_timepoint.time_zone.get_seconds()
         reference_timepoint.set_time_zone_to_utc()
-        reference_timepoint += Duration(seconds=adj_secs)
     return reference_timepoint + Duration(seconds=float(num_seconds))
 
 
--- a/metomi/isodatetime/data.py
+++ b/metomi/isodatetime/data.py
@@ -795,11 +795,12 @@ class TimePoint(object):
             (day_of_month, "day_of_month", None, int),
             (day_of_week, "day_of_week", None, int),
             (hour_of_day, "hour_of_day", None, int, float),
-            (hour_of_day_decimal, "hour_of_day_decimal", None, float),
+            (hour_of_day_decimal, "hour_of_day_decimal", None, int, float),
             (minute_of_hour, "minute_of_hour", None, int, float),
-            (minute_of_hour_decimal, "minute_of_hour_decimal", None, float),
+            (minute_of_hour_decimal, "minute_of_hour_decimal", None, int,
+             float),
             (second_of_minute, "second_of_minute", None, int, float),
-            (second_of_minute_decimal, "second_of_minute_decimal", None,
+            (second_of_minute_decimal, "second_of_minute_decimal", None, int,
              float),
             (time_zone_hour, "time_zone_hour", None, int),
             (time_zone_minute, "time_zone_minute", None, int)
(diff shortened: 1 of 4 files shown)


metomi/isodatetime/data.py
```diff
@@ -320,24 +320,35 @@ class Duration(object):
...
-    days (default 0): number of days in the duration
-    hours (default 0): number of hours in the duration
-    minutes (default 0): number of minutes in the duration
-    seconds (default 0): number of seconds in the duration
-    standardize (default False): boolean that, if True, switches on
-        adjusting the attributes so that small units have minimal values.
-        For example, 3664.4 seconds would become 1 hour, 1 minute, and 4.4
-        seconds. Attributes will not adjust for units that are inexact
-        (months and years).
+    years (int, default = 0): number of calendar years in the duration (an
```
oliver-sanders, line 335: I have no idea what format this docstring was supposed to be in before, but it was definitely invalid!

> **FYI:** We are trying to pull towards Google Style docstrings as these can be read by Napoleon (the package as opposed to the historical figure) which allows us to autodocument.

```suggestion
    years (int): number of calendar years in the duration (an
```

MetRonnie:
> not sure if users were able to use durations with float anywhere?

On master:
```python-console
>>> TimePoint(year=2020) + Duration(years=1.5)
2021-01-01T00:00:00Z
>>> TimePoint(year=2020) + Duration(months=1.5)
TypeError: 'float' object cannot be interpreted as an integer
>>> TimePoint(year=2020) + Duration(weeks=1.5)
2020-01-08T00:00:00Z
>>> TimePoint(year=2020) + Duration(days=1.5)
2020-01-02T00:00:00Z
```

metomi/isodatetime/datetimeoper.py
```diff
@@ -21,10 +21,8 @@
 import os
 import time
 
-from .data import (
-    Calendar,
-    get_timepoint_for_now as now2point,
-    get_timepoint_from_seconds_since_unix_epoch as seconds2point)
```
kinow, line 27: I think `seconds2point` not being used here any more, can be removed later? I think it should be used in tests, but I guess that can be removed too?

For later I guess.

PR #169, opened 2020-06-30
Prepare 2.0.2

MetRonnie:
It's my first time so hopefully I've done it right

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # ----------------------------------------------------------------------------
-# Copyright (C) 2013-2019 British Crown (Met Office) & Contributors.
+# Copyright (C) British Crown (Met Office) & Contributors.
 #
 # This program is free software: you can redistribute it and/or modify
 # it under the terms of the GNU Lesser General Public License as published by
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,11 +7,15 @@ for a full listing of issues for each release.
 
 ## isodatetime 2.0.2 (Upcoming, 2020)
 
-This is the 14th release of isodatetime.
-
-This release requires Python 3.5 or above.
+This is the 14th release of isodatetime. Requires Python 3.5+.
 
 ### Noteworthy Changes
+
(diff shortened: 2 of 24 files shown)


.codecov.yml
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 # ----------------------------------------------------------------------------
-# Copyright (C) 2013-2019 British Crown (Met Office) & Contributors.
+# Copyright (C) British Crown (Met Office) & Contributors.
 #
```
MetRonnie, line 4: Am I right that we are no longer including the year?

CHANGES.md
```diff
@@ -7,12 +7,16 @@ for a full listing of issues for each release.
 
 ## isodatetime 2.0.2 (Upcoming, 2020)
```
MetRonnie, line 9: Will need to update the date on release day

.codecov.yml
kinow: I think so. Only in one place now I think (License file? Can't recall)

oliver-sanders: Thanks for doing this, very observant.

You can follow the model of Cylc Flow, just add a line in the `README.md`.

See the Cylc Flow PR here: https://github.com/cylc/cylc-flow/pull/3583/files

MetRonnie: Added, let me know if it should be moved up or down

oliver-sanders:
Will need to bump the `__version__`.

.codecov.yml
oliver-sanders: Looks fine to me.

MetRonnie:
> Will need to bump the `__version__`.

Isn't that done? https://github.com/metomi/isodatetime/pull/169/files#diff-183a38a2f9f46f483668085b34665c9c

oliver-sanders:
Sorry, couldn't find it when I looked.

MetRonnie:
In that case I take it I don't need to wait for Tim?

oliver-sanders:
Two reviews sufficient. Go for it.

Issue #171, opened 2020-07-01
Automation of release process


MetRonnie:
This is just to brainstorm increasing the automation in releases.

Things that ought to be possible in GH Actions:
- [ ] Automatically set the version number in `__init__.py` from the release version (although I guess you lose out on the review process if you make a mistake with the release?)
- [ ] Automatically write the "released on" date in the changelog
- [x] Automatically update the copyright year
- [ ] Upload the build to PyPI (not sure about authentication)

This could be triggered by creating a GH Release instead of pushing a tag (GH release creates the tag instead), which is how it's done on [cylc/vscode-cylc](https://github.com/cylc/vscode-cylc) and [cylc/language-cylc](https://github.com/cylc/language-cylc)

Anyone have any comments or other ideas?

Update:
- [ ] Could we have it change milestones when an PR is merged earlier than expected, or comment to remind us?

MetRonnie:
We could use https://github.com/pypa/gh-action-pypi-publish

MetRonnie:
Question: how important is it that the commit that is tagged is verified? Usually the release tag is on a PR merge commit, which, performed on GitHub dot com is verified, but if you have GH Actions push a commit it isn't verified, and I assume you would have to set up some sort of GPG key thing for it to be...

If that makes any sense...

kinow:
I think that should be OK, but @hjoliver or @dpmatthews might know if there's a security requirement for the release commit to be signed.

oliver-sanders:
From previous discussion on Riot I think we should be happy to use the pypa GH action in combination with GH deploy keys for authentication.

> although I guess you lose out on the review process if you make a mistake with the release?

How about using a GH action to create a pull request with the changes (version number, copyright year, etc), then firing a deployment action off of that PR?

MetRonnie:
Something I just found out about GH actions: when you have
```yml
on:
  pull_request:
```
and you use `actions/checkout`, that doesn't exactly check out the PR branch - it actually creates a merge commit (PR branch into master (or whatever the base is)) and checks out that merge commit, locally on the runner. This is actually what Travis does as well.

MetRonnie:
Some features I wish Actions had, but doesn't:

1. The ability to trigger on a particular PR branch name. Currently you can only trigger off the *base* branch name using
    ```yaml
    on:
      pull_request:
        branches: [master]
    ```
    You can however use an `if` on the `job`,
    ```yaml
    jobs:
      check-release-pr:
[...]

PR #174, opened 2020-07-07
Automatically update copyright year every Jan 1st

MetRonnie:
These changes partially address #171 

This GitHub Actions workflow will create a PR with the updated copyright year in README.md at 12:00 UTC every new year's day.

This sort of workflow also serves as a proof-of-concept for other automation to be addressed from #171 

Demo:
- The workflow: https://github.com/MetRonnie/spork/runs/843244544?check_suite_focus=true
- The PR it created: https://github.com/MetRonnie/spork/pull/5

### DIFF
--- a/.github/workflows/update_copyright.yml
+++ b/.github/workflows/update_copyright.yml
@@ -1,0 +1,51 @@
+name: Update copyright year
+
+on:
+  schedule:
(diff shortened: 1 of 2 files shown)


MetRonnie:
Good point, at the very least this can act as a PoC

PR #181, opened 2020-08-28
GH Actions: Fix version checking in release workflow

### LINKED ISSUE #179: GH Actions: Release stage 1 version checking
In
https://github.com/metomi/isodatetime/blob/65d03e95cdcf303a54025853d7b02dd34dd4622d/.github/workflows/create_release_pr.yml#L84-L96

we should probably use Python's `packaging.version.Version()` for comparing versions, instead of just the strings, as things like `1.0` are implicitly the same as `1.0.0`

MetRonnie:
Use Python's `packaging.version.Version()` for comparing the `workflow_dispatch` inputted version with the ones on PyPI.org, instead of just comparing the strings, as things like `1.0` are implicitly the same as `1.0.0`.

Closes #179


kinow (approved):
Looks good to me. Tested as follows:

```bash
# installed packaging...
(venv) kinow@ranma:~/Development/python/workspace/isodatetime$ SETUP_PY_VERSION=$( python setup.py --version )
(venv) kinow@ranma:~/Development/python/workspace/isodatetime$ python .github/bin/cmp_py_versions "$SETUP_PY_VERSION" "$SETUP_PY_VERSION"
```

Then changed the version to an invalid value.

```diff
diff --git a/metomi/isodatetime/__init__.py b/metomi/isodatetime/__init__.py
index 362c702..f3fc935 100644
--- a/metomi/isodatetime/__init__.py
+++ b/metomi/isodatetime/__init__.py
@@ -17,4 +17,4 @@
 # ----------------------------------------------------------------------------
 """Python ISO 8601 date time parser and data model/manipulation utilities."""
 
-__version__ = "2.0.2"
+__version__ = "this is not a valid version"
```

And got.

```bash
(venv) kinow@ranma:~/Development/python/workspace/isodatetime$ SETUP_PY_VERSION=$( python setup.py --version )
/home/kinow/Development/python/workspace/isodatetime/venv/lib/python3.7/site-packages/setuptools/dist.py:483: UserWarning: The version specified ('1!this is not a valid version') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
(venv) kinow@ranma:~/Development/python/workspace/isodatetime$ python .github/bin/cmp_py_versions "$SETUP_PY_VERSION" "$SETUP_PY_VERSION"
  File ".github/bin/cmp_py_versions", line 29, in <module>
    ver1, ver2 = [packaging.version.Version(i) for i in sys.argv[1:3]]
  File ".github/bin/cmp_py_versions", line 29, in <listcomp>
    ver1, ver2 = [packaging.version.Version(i) for i in sys.argv[1:3]]
  File "/home/kinow/Development/python/workspace/isodatetime/venv/lib/python3.7/site-packages/packaging/version.py", line 277, in __init__
    raise InvalidVersion("Invalid version: '{0}'".format(version))
packaging.version.InvalidVersion: Invalid version: '1!this is not a valid version'

```

.github/bin/cmp_py_versions
```diff
@@ -0,0 +1,32 @@
...
+    You can also check if a single version number is valid by passing it twice.
+"""
+
+import sys
+import packaging.version
+
+ver1, ver2 = [packaging.version.Version(i) for i in sys.argv[1:3]]
+
+if ver1 != ver2:
+    print(f'{ver1} not equal to {ver2}')
```
oliver-sanders, line 32: Better to `exit(1)` rather than relying on stdout to communicate failure.

Not a biggie for this usage but worth noting that stdout/err can be quite fragile interfaces due to buffering (see the env var `PYTHONUNBUFFERED`), control characters, shell startup files, system messages and all sorts of lovely things.

oliver-sanders (approved):
Ah, sorry didn't mention that we use a difference licence for this project (actually forgot myself), good catch.

One comment but not a blocker, feel free to address or merge.

.github/bin/cmp_py_versions
MetRonnie: Only problem is that makes it more difficult to handle in the node script step of the workflow https://github.com/metomi/isodatetime/blob/2e5b4c474ad909079e4bcd4fa6b168bdcb70f354/.github/workflows/create_release_pr.yml#L94-L100

PR #185, opened 2020-09-18
Replace pycodestyle with flake8

wxtim:
Other projects related to this one have been moving to using flake8.
This change proposed for the sake of consistency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,11 +43,11 @@ install:
   # PyTest to use environment variables in its configuration (e.g. TZ=UTC).
   - pip install coverage pytest-cov pytest-env pytest-runner
   # Code style dependencies
-  - pip install pycodestyle
+  - pip install flake8
   - pip install -e .
 
 before_script:
-  - pycodestyle -v metomi/isodatetime
+  - flake8 -v metomi/isodatetime
 
 script:
   - |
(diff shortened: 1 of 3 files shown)


oliver-sanders (approved):

MetRonnie (approved):

MetRonnie:
Should we add flake8 to https://github.com/metomi/isodatetime/blob/fbe537d6ecdf10af8aac6fa415a13a8c1df384b4/setup.py#L98

oliver-sanders:
Yes but I let it slip since pyccodestyle wasn't there before.

If adding it there (good idea) add the tests_require to an `all` group so we can `pip install -e .[all]` as with other repos.

wxtim:
I have now modified setup.py to include flake8 and the option to use `pip install -e .[all]`

.travis.yml
```diff
@@ -43,11 +43,11 @@ install:
   # PyTest to use environment variables in its configuration (e.g. TZ=UTC).
   - pip install coverage pytest-cov pytest-env pytest-runner
   # Code style dependencies
-  - pip install pycodestyle
+  - pip install flake8
   - pip install -e .
```
oliver-sanders, line 47:
```suggestion
  - pip install -e .[all]
```

MetRonnie (approved):
I learnt a thing or two about how `pip install` works from this 👍

oliver-sanders:
Just waiting for tests to pass...

wxtim:
> I learnt a thing or two about how pip install works from this +1

Moi aussi!

PR #187, opened 2020-10-16
Refactor TimePoint getters and fix decimal quantity dump rounding

### LINKED ISSUE #180: Simplify TimePoint getters
There's no need to have this big block: https://github.com/metomi/isodatetime/blob/65d03e95cdcf303a54025853d7b02dd34dd4622d/metomi/isodatetime/data.py#L1178-L1257

when we can use the `@property` decorator. Besides, some of those properties in there are redundant, e.g. `hour_of_day`

### LINKED ISSUE #184: TimePoint dumping buggy for decimals that round up to 1
Example:
```python-console
>>> TimePoint(year=2020, second_of_minute=5, second_of_minute_decimal=0.9999999)
<metomi.isodatetime.data.TimePoint: 2020-01-01T00:00:05,1.Z>
```
Instead of `00:00:05,1.Z` it should be `00:00:06,0Z`.

That would make the behaviour in line with rounding down, e.g.
```python-console
>>> TimePoint(year=2020, second_of_minute=6, second_of_minute_decimal=0.000000000001)
<metomi.isodatetime.data.TimePoint: 2020-01-01T00:00:06,0Z>
```
Another question is if the stripping of trailing zeroes when rounding to 6 decimal places is misleading about the precision/number of significant figures

MetRonnie:
This branch was kicking around for a while; I probably meant to create a PR earlier.

- Closes #180 - use only `@property` decorators for getter methods, instead of having a mix of those and a `get(self, property_name)` method
- Closes #184 - fix bug where decimal quantities were getting rounded up to illegal values when dumping timepoints, e.g. minute of hour = `59.999999999` --> `60.0`. Just truncate the quantity instead (in this specific circumstance), so it dumps as `59.999999` to 6 d.p.
- Now prevents TimePoint from being init'd without a year (unless it's a truncated TP)

---
- [x] Change log updated
- [x] Unit tests included


metomi/isodatetime/data.py
```diff
@@ -880,11 +880,11 @@ class TimePoint:
 
     Keyword arguments (usually default to None if not provided):
 
-    expanded_year_digits (default 0) - an agreed-upon number of extra
-        digits to represent the year, beyond the default of 4. For example,
+    num_expanded_year_digits (default 0) - an agreed-upon number of extra
```
kinow, line 883: I don't have a preference for `expanded_year_digits` or `num_expanded_year_digits`, but as this property is changing for `TimePoint`, should we change it in the rest of the code too? `grep`'ing for `expanded.year.digit`, looks like `test__00.py`, `parser_spec.py`, `data.py`, `parsers.py`, `dumpers.py`, and `README.md` mention expanded year digits.

Are those reference OK, in which case this rename here is necessary as this property doesn't represent TimePoint's expanded year digits? (not sure if that makes sense, sorry :laughing: ), or are we going to rename them too later?

metomi/isodatetime/dumpers.py
```diff
@@ -61,6 +61,8 @@ class TimePointDumper(object):
 
     """
 
+    FLOAT_DECIMAL_PLACES = 6
```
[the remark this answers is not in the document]
MetRonnie: Good spot

metomi/isodatetime/data.py
MetRonnie: It's the difference between the number of digits and the actual digits themselves 

https://github.com/metomi/isodatetime/blob/2b908adb4745cc3e5ff8a16a62fbb9c75b4348d6/metomi/isodatetime/data.py#L1110-L1111

https://github.com/metomi/isodatetime/blob/2b908adb4745cc3e5ff8a16a62fbb9c75b4348d6/metomi/isodatetime/data.py#L1233-L1237

(not sure why those two are showing different indentations, they're the same)

Before, we had `TimePoint.expanded_year_digits` which was the number of digits, and `TimePoint.get("expanded_year_digits")` which was the digits themselves. The former should probably have always been called `num_expanded_year_digits`, whereas the latter has not been renamed (just converted to `@property`)

MetRonnie:
(The tests on Travis have passed, it's just Codacy complaining about an unassigned variable or something in a test)

metomi/isodatetime/tests/test_00.py
```diff
@@ -1084,9 +1093,6 @@ def test_timepoint_dumper(self):
                     num_expanded_year_digits=num_expanded_year_digits)
                 self.assertRaises(ctrl_exception, dumper.dump,
                                   ctrl_timepoint, format_)
-        value_error_timepoint = data.TimePoint(minute_of_hour=10)
-        value_error_timepoint._minute_of_hour = "1O"
-        self.assertRaises(ValueError, dumper.dump, value_error_timepoint, "%M")
```
oliver-sanders, line 1089: Behaviour change?

oliver-sanders (commented):
Looks fine, adds more breaking changes for Cylc to deal with but they are pretty small and easy to handle.

oliver-sanders:
BTW to make codacy happy just dedent that function and remove the unused `self` arg.

oliver-sanders (approved):

metomi/isodatetime/tests/test_00.py
MetRonnie: Now that TimePoints are "immutable", we shouldn't have to worry about mutating a property to an invalid value

MetRonnie:
Oh, codacy only shows the newest problem it has found with the code. Last time it was failing because of this [expression-not-assigned (W0106)](https://app.codacy.com/manual/metomi/isodatetime/file/49294204436/issues/source?bid=5063507&fileBranchId=5063507#l1452). Trouble is, if you assign it, flake8 will flag it as an unused variable!

MetRonnie:
@kinow just a reminder this is awaiting your review, no rush though

kinow (approved):

PR #193, opened 2021-07-30
Fix west-of-UTC time zone bug

### LINKED ISSUE #192: get_local_time_zone() broken for time zones with minutes west of UT
The timezone functions are not working properly for non-standard timezones behind UTC that have either:
- zero hours but non-zero minutes (e.g. -00:30)
- minutes that aren't 0 or 30 (e.g. -02:15)

#### Steps to reproduce

(Note: the `TZ` env var +/- convention is backwards, so that Eastern Standard Time is written `TZ=EST+05:00` instead of UTC-05:00)

```
$ TZ=XXX+02:15 python
>>> from metomi.isodatetime import timezone
>>> timezone.get_local_time_zone()
(-2, 45)  # should be (-2, 15) or (-2, -15)
>>> timezone.get_local_time_zone_format()
'-0245'  # should be '-0215'
```

```
$ TZ=XXX+00:30 python
>>> from metomi.isodatetime import timezone
>>> timezone.get_local_time_zone()
(0, 30)  # should be (0, -30)
>>> timezone.get_local_time_zone_format()
'+0030'  # should be '-0030'
```

#### Additional context

If the timezone is east of UTC, it works ok

```
$ TZ=XXX-02:15 python
>>> from metomi.isodatetime import timezone
>>> timezone.get_local_time_zone()
(2, 15)
>>> timezone.get_local_time_zone_format()
'+0215'
```

MetRonnie:
Closes #192

Fixes a bug where `metomi.isodatetime.timezone` would return incorrect results for certain non-standard/unusual system time zones. Also, `get_local_time_zone()` now always returns negative minutes for time zones behind UTC. 

This PR also drops support for Python 3.5, and makes `-m 'not slow'` the default for pytest.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python-version: ['3.5', '3.6', '3.7', '3.8', '3.9', '3.10-dev']
+        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10-dev']
         coverage: [false]
         include:
           # Modify existing configurations:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -50,7 +50,7 @@ jobs:
         env:
           TZ: ${{ matrix.tz }}
         run: |
-          PYTEST_ARGS=('-n' '2')
+          PYTEST_ARGS=(-n 2 -m 'slow or not slow')
           if ${{ matrix.coverage }}; then
             PYTEST_ARGS+=('--cov=metomi/isodatetime')
           fi
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,5 +1,3 @@
-# -*- coding: utf-8 -*-
-# ----------------------------------------------------------------------------
 # Copyright (C) British Crown (Met Office) & Contributors.
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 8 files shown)


pytest.ini
```diff
@@ -14,10 +12,16 @@
...
-# Some tests fail when running in a different time zone (e.g. Pacific/Auckland)
-markers = slow
-testpaths = metomi/isodatetime/tests
+addopts =
+    -v
+    -s
+    -ra
+    --color=auto
+    --doctest-glob='*.md'
+    -m 'not slow'
```
kinow, line 23: One job in GH Actions timed out after 45 minutes. Maybe there's something missing here?

kinow (approved):
Code and tests look good! Leaving one comment about a job that got stuck, but not related to the west-of-UTC time zone bug, not a blocker I think. Thanks!

```bash
(venv) kinow@ranma:~/Development/python/workspace/isodatetime$ TZ=EST+05:00 python
Python 3.8.3 (default, May 19 2020, 18:47:26) 
[GCC 7.3.0] :: Anaconda, Inc. on linux
>>> from metomi.isodatetime import timezone
>>> timezone.get_local_time_zone()
(-5, 0)
>>> 
```

pytest.ini
MetRonnie: I'm pretty sure it's just the coverage run being slow

.github/workflows/test.yml
```diff
@@ -50,7 +50,7 @@ jobs:
         env:
           TZ: ${{ matrix.tz }}
         run: |
-          PYTEST_ARGS=('-n' '2')
+          PYTEST_ARGS=(-n 2 -m 'slow or not slow')
```
oliver-sanders, line 53: Isn't `slow or not slow` equivalent to not specifying any marks?

oliver-sanders (approved):

.github/workflows/test.yml
MetRonnie: No because of https://github.com/metomi/isodatetime/blob/19929352802aa4c4fe9a72a0475c410775dc58bc/pytest.ini#L22

MetRonnie:
Merging with 2 approvals

PR #199, opened 2021-12-02
Pin to Python 3.9.

hjoliver:


MetRonnie (commented):
I don't think this is needed for isodatetime because pyuv isn't a dependency?

PR #202, opened 2022-01-09
Adding get_first_after convenience method

WaltWh:
Useful for when you're using TimeRecurrence to schedule tasks. Looping through "get_next" could cause performance problems if, say, you're trying to schedule something every second, or over extended periods of time.


WaltWh:
Updated formatting per flake8. Sorry, wasn't sure what linter to use.

WaltWh:
aaaand that's embarrassing. I have no idea where I got those values. Did the date math on WolframAlpha this time to make sure the values are correct.

oliver-sanders:
Hello there and thanks for your interest in isodatetime!

> say, you're trying to schedule something every second, or over extended periods of time

Yeah, that would be painfully slow.

Unfortunately whilst you can convert a duration into seconds, the result in seconds is not necessarily equal to the original duration, e.g. the length of a year varies due to leap years. This will cause the `get_first_after` interface to yield different results to a `get_next` loop for some recurrences:

```python
from metomi.isodatetime.data import *

start = TimePoint(year=2000)
duration = Duration(years=1)
rec = TimeRecurrence(start_point=start, duration=duration)

def get_next_method():
    point = start
    for _ in range(10):
        print(point)
        point = rec.get_next(point)

def get_first_after_method():
    point = start
    for _ in range(10):
        print(point)
        point = rec.get_first_after(point)

print('# get_next')
get_next_method()
print('# get_first_after')
get_first_after_method()
```

```console
$ python test.py
# get_next
2000-01-01T00:00:00Z
2001-01-01T00:00:00Z
2002-01-01T00:00:00Z
2003-01-01T00:00:00Z
2004-01-01T00:00:00Z
2005-01-01T00:00:00Z
2006-01-01T00:00:00Z
2007-01-01T00:00:00Z
2008-01-01T00:00:00Z
2009-01-01T00:00:00Z
# get_first_after
2000-01-01T00:00:00Z
2001-01-01T00:00:00Z
2001-12-31T00:00:00Z
2002-12-31T00:00:00Z
2003-12-31T00:00:00Z
2004-12-31T00:00:00Z
2005-12-30T00:00:00Z
2006-12-30T00:00:00Z
2007-12-30T00:00:00Z
2008-12-30T00:00:00Z
```

This is why we count from the start point and push calculations through the `TimePoint._tick_over` interface, which is annoying and really slow but the only way do handle all cases reliably.

The `get_first_after` approach is perfectly valid for many calculations e.g. `Duration(seconds=5)` but not for others e.g. `Duration(months=5)`. I'm be interested in the idea of detecting when this method is valid and automatically switching to the more efficient method when we are able to, falling back to the slower method only when we have to.

I'm also interested in trying out [numba](https://numba.pydata.org) on this library to see if it can give us a performance boost for things like `TimePoint._tick_over`.

MetRonnie:
> The get_first_after approach is perfectly valid for many calculations e.g. Duration(seconds=5) but not for others e.g. Duration(months=5). I'm be interested in the idea of detecting when this method is valid and automatically switching to the more efficient method when we are able to, falling back to the slower method only when we have to.

It's possible to use `Duration.is_exact()` to check whether the method is valid

MetRonnie:
> The get_first_after approach is perfectly valid for many calculations e.g. Duration(seconds=5) but not for others e.g. Duration(months=5). I'm be interested in the idea of detecting when this method is valid and automatically switching to the more efficient method when we are able to, falling back to the slower method only when we have to.

It should be possible to apply the method if `self.duration.is_exact() == True`

WaltWh:
That totally makes sense. I'll see if I can dig into it a little deeper when I have the chance.

> It's possible to use Duration.is_exact() to check whether the method is valid

Awesome, I'll try branching on that and do some testing. Looks like the results of `TimePoint.__sub__` are always exact, so we'll only need to switch over if the period itself has a month or year unit. In practical usage I'd expect that to mean that even in the most extreme-but-plausible use case I can think of, someone setting a bimonthly task as `R/0001-01-01/P0.5M`, the loop would never be expected to repeat more than ~2000 * 12 * 2 ~= 50000 times in one call. That's a little much but it should be manageable, at least compared to the ~64800000000 you'd end up with if they used a similar method to count by seconds.

That's not counting e.g. a "P0.001M" duration but I feel like if someone's getting into that sort of edge case they can reasonably expect to have to actually think about what they're doing.


All that to say, yes, sounds good, I'll make the change, and make sure to add test cases that cover it.


> This is why we count from the start point and push calculations through the TimePoint._tick_over interface, which is annoying and really slow but the only way do handle all cases reliably.

Is there already a convenience method somewhere with a similar interface to what I added, to automatically do this? I didn't find anything when I checked (and in fact was rather frustrated at the fact that `get_next` didn't do just that when given a non-member timepoint).

MetRonnie:
>  the most extreme-but-plausible use case I can think of, someone setting a bimonthly task as R/0001-01-01/P0.5M

It isn't possible to use decimals on date units (only time i.e. hours, mins, secs)

metomi/isodatetime/data.py
```diff
@@ -312,11 +312,20 @@ def get_prev(self, timepoint: "TimePoint") -> "TimePoint":
...
+                    (timepoint - self.start_point).get_seconds(),
+                    self.duration.get_seconds())
+                return timepoint + (self.duration - Duration(
+                    seconds=floor(seconds_since)))
+            else:
+                # Since duration is inexact, we have to iterate
+                current = self.start_point
+                while current <= timepoint:
+                    print(current)
+                    current = self.get_next(current)
```
MetRonnie, line 327: Two things:
- `.get_next()` can return `None`, so need to avoid type error
- don't print
```suggestion
                while current is not None and current <= timepoint:
                    current = self.get_next(current)
```

metomi/isodatetime/data.py
```diff
@@ -309,6 +310,26 @@ def get_prev(self, timepoint: "TimePoint") -> "TimePoint":
             return prev_timepoint
         return None
 
+    def get_first_after(self, timepoint):
```
MetRonnie, line 313:
```suggestion
    def get_first_after(self, timepoint):
        """Return the next timepoint in the series after the given timepoint
        which is not necessarily part of the series.
        
        If the given timepoint is before the start point, return the
        start point, or if it is after the end point, return None.
        """
```

metomi/isodatetime/data.py
```diff
@@ -309,6 +310,26 @@ def get_prev(self, timepoint: "TimePoint") -> "TimePoint":
...
+        if self._get_is_in_bounds(timepoint):
+            if self.duration.is_exact():
+                # Since it's exact, we can do maths instead of iterating
+                iterations, seconds_since = divmod(
+                    (timepoint - self.start_point).get_seconds(),
+                    self.duration.get_seconds())
+                return timepoint + (self.duration - Duration(
```
MetRonnie, line 320:
```suggestion
                    (timepoint - self._start_point).get_seconds(),
                    self._duration.get_seconds())
                return timepoint + (self._duration - Duration(
```

metomi/isodatetime/data.py
```diff
@@ -309,6 +310,26 @@ def get_prev(self, timepoint: "TimePoint") -> "TimePoint":
...
+                    seconds=floor(seconds_since)))
+            else:
+                # Since duration is inexact, we have to iterate
+                current = self.start_point
```
MetRonnie, line 324:
```suggestion
                current = self._start_point
```

metomi/isodatetime/data.py
```diff
@@ -309,6 +310,26 @@ def get_prev(self, timepoint: "TimePoint") -> "TimePoint":
...
+                while current <= timepoint:
+                    print(current)
+                    current = self.get_next(current)
+                return current
+        elif timepoint < self.start_point:
+            return self.start_point
```
MetRonnie, line 330:
```suggestion
        elif timepoint < self._start_point:
            return self._start_point
```

metomi/isodatetime/tests/test_00.py
```diff
@@ -894,6 +894,24 @@ def get_timerecurrence_membership_tests():
...
+          ("2020-01-02T00:00:00Z", "2020-01-05T00:00:00Z"),
+          ("2020-01-15T00:00:00Z", "2020-01-17T00:00:00Z"),
+          ("2021-01-01T00:00:00Z", "None")]),
+        ("R3/P700D/1961-W01-1T06,5Z",
+         [("1950-01-01T00:00:00Z", "1957-W10-1T06,5Z"),
+          ("1958-01-01T00:00:00Z", "1959-02-02T06:30:00Z"),
+          ("2021-01-01T00:00:00Z", "None")]),
+        ("R/2000-01-01T00:00:00Z/P1Y",
+         [("2001-01-01T00:00:00Z", "2002-01-01T00:00:00Z"),
+          ("2002-01-01T00:00:00Z", "2003-01-01T00:00:00Z")])
```
MetRonnie, line 911: Test for the type error condition I mentioned above
```suggestion
          ("2002-01-01T00:00:00Z", "2003-01-01T00:00:00Z")]),
        ("R1/2000-01-01T00:00:00Z/P1Y",
         [("2000-01-01T00:00:00Z", "None")])
```

MetRonnie (changes requested):
Another thing: it's more performant to access the private attributes, so I've suggested adding leading underscores:

metomi/isodatetime/data.py
WaltWh: get_next shouldn't ever return None since we've already checked that we're within the valid range. Probably doesn't hurt to confirm it though.

WaltWh: Oh, it can if we're sitting directly on the very last instant. got it

metomi/isodatetime/data.py
```diff
@@ -309,6 +310,31 @@ def get_prev(self, timepoint: "TimePoint") -> "TimePoint":
...
 
+    def get_first_after(self, timepoint):
+        """Return the next timepoint in the series after the given timepoint
+        which is not necessarily part of the series.
+        
+        If the given timepoint is before the start point, return the
+        start point, or if it is after the end point, return None.
+        """
+        if self._get_is_in_bounds(timepoint):
+            if self.duration.is_exact():
```
MetRonnie, line 321: Oops, I missed one suggestion I had meant to include. `self.duration` is `None` if `self.repetitions == 1` (because that means there is only a single point in the recurrence)
```suggestion
            if self._duration is not None and self._duration.is_exact():
```

MetRonnie (changes requested):

metomi/isodatetime/data.py
WaltWh: Rolled this suggestion into another change to fix formatting

MetRonnie (approved):
Thanks for the contribution!

oliver-sanders (approved):
Thanks, well put this into the next release.

WaltWh:
Glad I could contribute, this library has been super helpful for several of my projects.

Out of curiosity, is there any plan for when that release will be?

MetRonnie:
There isn't a "plan" as such 😛 We were expecting to release within the next couple of months, but nothing to stop us from doing a small release if you want us to

WaltWh:
No rush, was just curious. My dependent project will probably be releasing within the next few days (depending on whether another dependency releases on schedule) but I have no problem with having it install from github rather than pip for now.

WaltWh:
Thank you! Glad I could contribute, and glad my use case was reasonable enough to include.

PR #203, opened 2022-03-02
Tidy

MetRonnie:
Accumulation of various acts of tidying over the months.

One relatively noteworthy behaviour change: `TimePoint.seconds_since_unix_epoch` is now an `int` instead of `str`.


MetRonnie:
Superseded by #234

PR #204, opened 2022-03-29
cli: fix issue with negative durations

oliver-sanders:
Will close https://github.com/metomi/rose/issues/2569 in combination with a dependency change.

This is more an inconsistency than a bug from the perspective of isodatetime, however, from the Rose side it is a bug as we had previously worked to handle this case.

Before:
```console
$ isodatetime 2000 -s -P1D
isodatetime: error: argument --offset1/--offset/-s/-1: expected one argument
$ isodatetime 2000 -s=-P1D
1999
$ isodatetime 2000 -s \-P1D
isodatetime: error: argument --offset1/--offset/-s/-1: expected one argument
$ isodatetime 2000 -s=\-P1D
1999
```

After:
```console
$ isodatetime 2000 -s -P1D
1999
$ isodatetime 2000 -s=-P1D
1999
$ isodatetime 2000 -s \-P1D
1999
$ isodatetime 2000 -s=\-P1D
1999
```

(note, swap `isodatetime` for `rose date` and you will get the same result)

A bit crude but I think this is ok as no other arguments should start with the string `-P`.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -58,7 +58,7 @@ coverage:
 
 # files to ignore
 ignore:
-  - "isodatetime/tests*"
+  - "metomi/isodatetime/tests*"
 
 # turn off comments to pull requests
 comment: off
--- a/metomi/isodatetime/main.py
+++ b/metomi/isodatetime/main.py
@@ -174,7 +174,7 @@ from . import __version__
 from .datetimeoper import DateTimeOperator
 
 
-def parse_args():
+def parse_args(sys_args=None):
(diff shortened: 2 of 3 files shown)


wxtim:
See https://github.com/oliver-sanders/isodatetime/pull/2

MetRonnie (approved):

wxtim (approved):
- [x] Looks sane
- [x] Tests seem to test what they claim to test
- [x] Manually tested

PR #205, opened 2022-03-30
Change next release from 2.1 to 3.0

MetRonnie:
I propose changing the upcoming release from 2.1 to 3.0 due to the breaking changes:
- Immutability means you can no longer do things like `my_timepoint.year += 1`
- Replacement of `my_timepoint.get()` with properties
- #183 is sort of like a breaking change

I don't see any reason not to make this a major release. Considering we have been bitten by packages not obeying semver recently (e.g. markupsafe), I think it would be more responsible to make this a major release (even if few people use isodatetime outside of Cylc)

If this is merged I will update the milestone accordingly

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,9 +11,9 @@ ones in. -->
 
 --------------------------------------------------------------------------------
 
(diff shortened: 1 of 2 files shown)


oliver-sanders (approved):
A little dramatic for the scale of changes but ok.

PR #207, opened 2022-03-31
Fix version of `gh-action-pypi-publish` and add dependabot.yml

MetRonnie:
`gh-action-pypi-publish` does not have a `v1` tag like other actions, it was my bad to do that in #200

Good news is we can rely on dependabot to update the version from now on

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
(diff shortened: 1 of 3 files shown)


oliver-sanders (approved):

oliver-sanders:
> Good news is we can rely on dependabot to update the version from now on

:+1:

oliver-sanders:
LGTM

PR #231, opened 2023-05-22
Fix outdated CLI doc

### LINKED ISSUE #218: Remove reference to `-c` option in CLI help
The `isodatetime` CLI command is mostly a drop-in replacement for `rose date`, but the exception is the `-c` option which is not present in `isodatetime`. However, the CLI help text was copied over and still has references to `-c` that need to be removed.

See also https://github.com/metomi/rose/issues/2617

MetRonnie:
Closes #218


wxtim (approved):
Legit - I'd be happy to say this is a 1-review PR - If you agree, merge?

PR #239, opened 2023-10-18
added flake8-type-checking

markgrahamdawson:
Closes https://github.com/cylc/cylc-admin/issues/182

Added flake8-type-checking library to project

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,7 @@ below:
 * Mel Hall (Met Office, UK)
 * Ronnie Dutta (Met Office, UK)
 * Walt Whiteside (US)
+* Mark Dawson (Met Office, UK)
 
 (All contributors are identifiable with email addresses in the version control
 logs or otherwise.)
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -18,7 +18,10 @@
 
 [aliases]
 # so that running python setup.py test invokes pytest
-test=pytest
+test=
+    pytest
+    flake8
+    flake8-type-checking; python_version > "3.7"
```
MetRonnie, line 24: This isn't actually adding `flake8-type-checking` as a dependency. In Isodatetime we still haven't switched from `setup.py` to `setup.cfg` properly. And in fact it would make more sense to switch to `pyproject.toml` at this point

MetRonnie:
Marking as draft until #240 goes in

MetRonnie:
@markgrahamdawson If you retry this now it should be clear where to add it in `pyproject.toml`

MetRonnie (approved):

PR #240, opened 2023-10-18
Switch to `pyproject.toml`

MetRonnie:
Move from `setup.py` to `pyproject.toml`

Docs:
- https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html
- https://packaging.python.org/en/latest/specifications/declaring-project-metadata/

Also:
- Support Python 3.12
- GH Actions improvements


.github/workflows/test.yml
```diff
@@ -38,27 +48,48 @@ jobs:
...
+        run: pytest
 
-      - name: Upload coverage report
+      - name: Coverage report
         if: matrix.coverage
         run: |
-          coverage xml --ignore-errors
-          bash <(curl -s https://codecov.io/bash)
+          coverage xml
+          coverage report
```
oliver-sanders, line 66: Do we need this bit?

oliver-sanders (approved):

oliver-sanders:
It might be worth taking the opportunity to roll the pytest.ini and .bandit files into this.

.github/workflows/test.yml
MetRonnie: `coverage xml` - I think this is needed to convert the data into a compatible format for Codecov? But not sure.

`coverage report` - Not strictly needed but can be useful to see the coverage stats sometimes

MetRonnie:
> (& .coveragerc maybe?).

Decided against this as coverage configs can be quite lengthy and also the exclusion regexes would need extra escaping in TOML

wxtim (approved):
- [x] Looks good
- [x] Tried it locally with a couple of different pythons

PR #269, opened 2025-09-02
Bump min Python to 3.9

MetRonnie:


pyproject.toml
```diff
@@ -24,22 +24,20 @@ authors = [
     {name = "Met Office", email = "metomi@metoffice.gov.uk"}
 ]
 description = "Python ISO 8601 date time parser and data model/manipulation utilities"
-license = {text = "LGPLv3"}
+license = "LGPL-3.0-only"
```
MetRonnie, line 27: >Please use a simple string containing a SPDX expression for `project.license`. You can also use `project.license-files`. (Both options available on setuptools>=77.0.0).
>
>By 2026-Feb-18, you need to update your project and remove deprecated calls
>or your builds will no longer be supported.
>
>See https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license for details.

PR #284, opened 2026-04-10
setup: sync pyproject config from cylc-flow

oliver-sanders:


pyproject.toml
```diff
@@ -96,3 +96,34 @@ testpaths = [
 
 [tool.bandit]
 exclude_dirs = ["metomi/isodatetime/tests"]
+
+
+# Not mandated to use these tools, but if you do:
+
+[tool.ruff]
+line-length = 79
+target-version = "py312"
```
MetRonnie:
```suggestion
target-version = "py310"
```

pyproject.toml
```diff
@@ -96,3 +96,34 @@ testpaths = [
...
+
+[tool.ruff.format]
+quote-style = "preserve"
+
+
+[tool.black]
+line-length = 79
+target-version = ['py312']
```
MetRonnie:
```suggestion
target-version = ['py310']
```

MetRonnie (changes requested):

pyproject.toml
oliver-sanders: Good point!

MetRonnie (approved):

PR #285, opened 2026-05-19
optimise TimePoint._cmp

oliver-sanders:
Partially addresses https://github.com/cylc/cylc-flow/issues/7267

The `TimePoint.get_props` method is getting hammered in this example. This is causing `_copy` to get hammered in turn.

Much of this is due to `TimePoint._cmp`, but it's not clear that `_copy` is necessary for these comparisons.


MetRonnie:
Superseded by #288

PR #288, opened 2026-06-29
Efficiency: remove unnecessary copy operation

MetRonnie:
Alternative to https://github.com/metomi/isodatetime/pull/285

### DIFF
--- a/metomi/isodatetime/data.py
+++ b/metomi/isodatetime/data.py
@@ -23,6 +23,7 @@ from functools import lru_cache
 from math import floor
 import operator
 from typing import (
+    Any,
     Literal,
     cast,
     overload,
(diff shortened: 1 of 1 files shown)


metomi/isodatetime/data.py
```diff
@@ -1719,24 +1721,20 @@ def __add__(self, other: object) -> 'TimePoint':
...
         return new_timepoint
 
-    def get_props(self) -> list:
+    def get_props(self) -> list[tuple[str, Any]]:
         """Return the data properties of this TimePoint as a list of tuples."""
-        props = []
-        for attr in self.__slots__:
-            value = getattr(self, attr, None)
-            if callable(getattr(value, "_copy", None)):
-                value = value._copy()
```
MetRonnie, line 1736: I am convinced this copy was unnecessary. Note it's the time zone that was being copied. The only uses of this method:
- In the comparison function, where no references to the time zone are created
- In `def get_timepoint_properties_from_seconds_since_unix_epoch()`, where a local reference to the time zone is created, but not mutated

oliver-sanders: I think you are right.

The only mutable property returned is the time zone. This is a `TimeZone(Duration)` object which is `__hash__`able. It provides no methods that would mutate the object, you would have to meddle with a `_` prefixed variable to mess it up.

Seems reasonable.

oliver-sanders (approved):
This looks good to me, but we don't actually have any tests for this method.

